PR #2, opened 2021-11-05
Update README File and Shell script changes

snrssc:
Following are list of changes

README file - updated execution commands
start.sh - Modified the way base directory is identified. Fix for running with . start.sh
- Removed set -x and added echo command


ajayydv (approved):
LGTM +1

PR #4, opened 2021-11-05
Updated GCStoBQ to accept Parquet

jjaladi:


jjaladi:
Added logic to parse Parquet files

PR #27, opened 2021-11-26
Example of options classes, or generic templates

tims:
This is not yet for merging, just to spark a discussion on how we might improve our options/properties handling.

This approach uses the apache BeanUtils and the standard java validation framework for validating options pojos, this greatly reduces the amount of setup code.
eg: this duplicated per property:
```
prop = getProperties().getProperty(PROP);
if (StringUtils.isAllBlank(prop)) {
LOGGER.error("{} is required parameter. ", PROP);
throw new IllegalArgumentException(...)
}
LOGGER.info(PROP, prop);
etc..
```

Becomes this:
```
GCSInputOptions inputOptions = optionsFactory.as(GCSInputOptions.class).create();
```

This is only for options that have values passed into the template, other TemplateConstants still need to be kept somewhere (eg, the GCS "header", "inferSchema", "com.databricks.spark.avro" constants).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0"
-         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
   <groupId>com.google.cloud</groupId>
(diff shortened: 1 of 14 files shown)


tims:
So... I just pushed a generic template to this branch, for an extreme comparison.
Rather than write a pipeline for each input output pair, this template can mix and match any reader and writer pair.
It works with anything that only uses format, options and (optionally) path. 
See `com.google.cloud.dataproc.templates.GenericTemplate`
 
Parquet to CSV
```
input.format=parquet
input.parquet.path=gs://yadavaja-sandbox-tsell-us/output/shakespeare/parquet/
output.format=csv
output.csv.path=gs://yadavaja-sandbox-tsell-us/output/test/csv/
output.csv.header=true
```

BigQuery to BigQuery
```
input.format=bigquery
input.bigquery.table=project:dataset.table1
output.format=bigquery
output.bigquery.table=project:dataset.table2
```

GCS Avro to Spanner (this almost works, we just need to register the custom spanner dialect in my other PR)
```
input.format=avro
input.avro.format=gs://bucket/output/avro/
output.format=jdbc
output.jdbc.url=jdbc:cloudspanner:/projects/myprojectid/instances/myinstanceid/databases/mydatabaseid?lenient=true
output.jdbc.dbtable=mytablename
output.jdbc.driver=com.google.cloud.spanner.jdbc.JdbcDriver
output.jdbc.createTableOptions=PRIMARY KEY (id)
output.jdbc.isolationLevel=NONE
```

tims:
closing without merging

tims:
See related issue #28

PR #29, opened 2021-12-01
General template

tims:
Create GeneralTemplate for input/query/output templates configured via yaml
Has its own main method.

Example config that reads a public bigquery dataset, queries it in spark and writes to local output (can also write to gcs).

To test locally if you have spark installed:
```
spark-submit --master local \
    --conf spark.driver.extraJavaOptions='-Dlog4j.configuration=file:my/local/log4j.properties' \
    --class com.google.cloud.dataproc.templates.GeneralTemplate \
    target/dataproc-templates-1.0-SNAPSHOT.jar \
    --config config.yaml
```

For testing on dataproc batch, upload jars and config to `${GCS_STAGING_BUCKET}` as per `start.sh` and change the class, and add a files flag to copy the config.yaml to the working dir so we can pass it in as a parameter:
```
gcloud beta dataproc batches submit spark \
  --project=${GCP_PROJECT} \
  --region=us-west1 --subnet ${SUBNET} \
  --jars=file:///usr/lib/spark/external/spark-avro.jar,${GCS_STAGING_BUCKET}/grpc-grpclb-1.40.1.jar,${GCS_STAGING_BUCKET}/${JAR_FILE} \
  --labels job_type=dataproc_template --deps-bucket=${GCS_STAGING_BUCKET}/ \
  --class com.google.cloud.dataproc.templates.GeneralTemplate \
  --files=${GCS_STAGING_BUCKET}/config.yaml \
  -- --config config.yaml
```

```
input:
  shakespeare:
    format: bigquery
    options:
      table: "bigquery-public-data.samples.shakespeare"
query:
  wordcount:
    sql: "SELECT word, sum(word_count) cnt FROM shakespeare GROUP by word ORDER BY cnt DESC"
  wordspercorpus:
    sql: "SELECT corpus, count(*) distictwords, sum(word_count) numwords FROM shakespeare GROUP by corpus ORDER BY numwords DESC"
output:
  wordcount:
    format: csv
    options:
      header: true
    path: output/wordcount/
    mode: Overwrite
  wordspercorpus:
    format: json
    path: output/wordspercorpus/
    mode: Overwrite
```


tims:
For testing this template, see: https://github.com/GoogleCloudPlatform/dataproc-templates/blob/e1971d919405623749ab72d1619547b596af7871/bin/start-general.sh

tims:
It does a few things that are very different to the current approaches that I hope you'll forgive while considering the template it self. 

* It has it's own main function so that it has access to the cli arguments.
* It uses a yaml file to configure itself instead of properties
* It's much more flexible, and can be configured to operate on multiple sources and sinks at once (and can run sql on them)
* It does not catch all spark errors like the other templates do, it instead bubbles them up so the job can fail. It uses a finally block instead to politely shutdown the spark session. I think all the templates should do this, but was waiting and seeing how people respond to this before proposing that.
* It uses jakarta.validation to validate its configuration objects. This introduces some dependencies, but I think it's worth it. eg: Set<ConstraintViolation<GeneralTemplateConfig>> violations = config.validate(). This catches things like missing required fields, but also more specific things like declaring the output of a sql query with a conflicting name to an input. Or declaring an output table but it doesn't match any input or a query. Basically it's a java standard way of giving good error messages about pojos.

src/main/java/com/google/cloud/dataproc/templates/config/GeneralTemplateConfig.java
```diff
@@ -0,0 +1,157 @@
...
+  public @interface GeneralTemplateConfigAnnotation {
+
+    String message() default "inconsistent config";
+
+    Class<?>[] groups() default {};
+
+    Class<? extends Payload>[] payload() default {};
+  }
+
+  public static class GeneralTemplateConfigValidator
```
tims, line 116: This is custom annotation validator, the `@NonEmpty` `@NotNull` constraints are checked regardless (and in the child pojos, this validator checks that the keys in the input, query and output maps don't conflict or match as required.

src/main/java/com/google/cloud/dataproc/templates/config/GeneralTemplateConfig.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.Set;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@GeneralTemplateConfigAnnotation
+public class GeneralTemplateConfig {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(GeneralTemplateConfig.class);
+
+  @NotEmpty() private Map<String, @Valid InputConfig> input = ImmutableMap.of();
```
tims, line 48: The `@Valid` annotation ensures any of these instances are also validated. Not just this field's `@NotEmpty` constraint.

src/main/java/com/google/cloud/dataproc/templates/config/InputConfig.java
```diff
@@ -0,0 +1,62 @@
...
+
+import com.google.common.base.MoreObjects;
+import com.google.common.collect.ImmutableMap;
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+import java.util.Map;
+
+public class InputConfig {
+
+  @NotBlank private String format;
```
tims, line 26: We could make the format have a regex constraint on a fixed set of formats we will accept, but leaving it free text makes it more flexible.

src/main/java/com/google/cloud/dataproc/templates/config/OutputConfig.java
```diff
@@ -0,0 +1,75 @@
...
+import jakarta.validation.constraints.Pattern;
+import java.util.Map;
+
+public class OutputConfig {
+
+  @NotBlank private String format;
+  private String path;
+  @NotNull private Map<String, String> options = ImmutableMap.of();;
+
+  @Pattern(regexp = "Overwrite|ErrorIfExists|Append|Ignore")
```
tims, line 31: Corresponds to the [SaveMode](https://spark.apache.org/docs/latest/api/java/index.html?org/apache/spark/sql/SaveMode.html) enum

src/main/java/com/google/cloud/dataproc/templates/config/QueryConfig.java
```diff
@@ -0,0 +1,37 @@
...
+ * the License.
+ */
+package com.google.cloud.dataproc.templates.config;
+
+import com.google.common.base.MoreObjects;
+import jakarta.validation.constraints.NotBlank;
+
+public class QueryConfig {
+
+  @NotBlank private String sql;
```
tims, line 23: It would be nice to have a way to ask spark if this is valid sql before we start the spark session so we can fail faster, but I don't think there's an easy way to do that.

src/test/resources/config-example.yaml
```diff
@@ -0,0 +1,18 @@
...
+      SELECT
+      corpus, count(*) distictwords, sum(word_count) numwords
+      FROM shakespeare
+      GROUP by corpus
+      ORDER BY numwords DESC
+output:
+  wordspercorpus:
+    format: avro
+    path: gs://yadavaja-sandbox-tsell-us/output/wordspercorpus/
+    mode: Overwrite
```
tims, line 18: This example isn't required by any of the unit tests, I include it to make trying it easier without having to craft one version. We can remove this later.

src/test/resources/log4j-custom.properties
```diff
@@ -0,0 +1,5 @@
+# Set everything to be logged to the console
+log4j.rootCategory=INFO, console
+log4j.appender.console=org.apache.log4j.ConsoleAppender
+log4j.appender.console.layout=org.apache.log4j.PatternLayout
+log4j.appender.console.layout.ConversionPattern=%d{yy/MM/dd HH:mm:ss} %p %c{1}: %m%n
```
tims, line 5: We must pass a custom log4j file if we want to see application logs when using serverless spark. The current `start.sh` passes one that spark actually fails to load, which triggers it to load its package default config. This causes spark to be able to log from its packages, but applicatoin logs (com.google.etc..) are ignored.

This file is used by the `start-general.sh`, it lives in the test folder, so is not bundled with the jar. 
The script basically does the following:
```
gsutil cp log4j-custom.properties "gs://yadavaja-sandbox-tsell-us/config/log4j-custom.properties"
```
Then in the submit spark command it has:
```
  --properties=spark.driver.extraJavaOptions='-Dlog4j.debug=true -Dlog4j.configuration=file:log4j-custom.properties' \
  --files=gs://yadavaja-sandbox-tsell-us/config/log4j-custom.properties \
```
The `--files` flag downloads it to all locations in the working dir and the `log4j.configuration=file:log4j-custom.properties` property tells it to load from the working dir.

If you also add a debug flag to extraJavaOptions `-Dlog4j.debug=true`, then at the start of the logs you will see:
```
log4j: Using URL [file:log4j-custom.properties] for automatic log4j configuration.
log4j: Reading configuration from URL file:log4j-custom.properties
... etc
log4j: Finished configuring.
```

After that you can now see the application logs, eg:
```
21/12/02 08:03:46 INFO GeneralTemplate: Parsing arguments [--config, config.yaml]
21/12/02 08:03:46 INFO GeneralTemplate: Loading config from file:///tmp/srvls-batch-f4c0cd26-2e2e-4fdb-841d-c021d0fb1e04/config.yaml
21/12/02 08:03:46 INFO GeneralTemplate: File fetched
21/12/02 08:03:46 INFO GeneralTemplate: Parsing yaml
21/12/02 08:03:47 INFO GeneralTemplate: Config loaded
GeneralTemplateConfig{input={shakespeare=InputConfig{format=bigquery, options={table=bigquery-public-data.samples.shakespeare}, path=null}}, output={wordspercorpus=OutputConfig{format=avro, path=gs://yadavaja-sandbox-tsell-us/output/wordspercorpus/, options={}, mode=Overwrite}}, query={wordspercorpus=QueryConfig{sql=SELECT corpus, count(*) distictwords, sum(word_count) numwords FROM shakespeare GROUP by corpus ORDER BY numwords DESC
}}}
```

franklinWhaite:
> It does a few things that are very different to the current approaches that I hope you'll forgive while considering the template it self.
> 
> * It has its own main class `com.google.cloud.dataproc.templates.GeneralTemplate` so that it has access to the cli arguments.
> * It has its own `start-general.sh` start script as it uses a different main class and needs to pass in the config file. This is much simpler than the existing start script, for illustration purposes, not final.
> * It's much more flexible, and can be configured to operate on multiple sources and sinks at once (and can run sql on them)
> * It does not catch all spark errors like the other templates do, it instead bubbles them up so the job can fail. It uses a finally block instead to politely shutdown the spark session. I think all the templates should do this, but was waiting and seeing how people respond to this before proposing that.
> * it does not define properties very every kind of configuration per format, it instead relies on the user providing the format options and path for the inputs and outputs that conform the the spark documentation for the DataFrameReader and DataFrameWriter they wish to use. This allows it to support _any_ input and output that uses them.
> * It uses a yaml file to configure itself instead of using properties bundled with the jar. Passed in via command line argument `--config config.yaml`. It introduces configuration pojos under `com.google.cloud.dataproc.templates.config` which are used to parse the yaml.
> * It uses jakarta.validation to validate its configuration objects. This introduces some dependencies, but I think it's worth it. eg: `Set<ConstraintViolation<GeneralTemplateConfig>> violations = config.validate()`. This catches things like missing required fields, but also more specific things like declaring the output of a sql query with a conflicting name to an input. Or declaring an output table but it doesn't match any input or a query. Basically it's a java standard way of giving good error messages about pojos.

1. Having several main classes might be confusing for users. Might be more user friendly to have this general template called from the base template, i see this is already supported, maybe this should be the default way to run the generic template.
2. The flexibility of configuring several sources and syncs is great, we should eventually document the sources and syncs supported by the general template.
3. I think that it’s great that SQL can be provided! 
4. I agree that relying on the user providing the format is a good option 
5. It's awesome that we now have a way to overwrite properties at runtime with the .yaml file

src/main/java/com/google/cloud/dataproc/templates/config/InputConfig.java
franklinWhaite: Agree with flexibility

src/main/java/com/google/cloud/dataproc/templates/config/QueryConfig.java
franklinWhaite: Agree that this is desirable, but not urgent. we could create a FR to not loose track of this

src/test/resources/log4j-custom.properties
franklinWhaite: @snrssc thoughts on this?

src/test/resources/config-example.yaml
franklinWhaite: agree, should remove before merging

tims:
> * Having several main classes might be confusing for users. Might be more user friendly to have this general template called from the base template, i see this is already supported, maybe this should be the default way to run the generic template.
> * The flexibility of configuring several sources and syncs is great, we should eventually document the sources and syncs supported by the general template.
> * I think that it’s great that SQL can be provided!
> * I agree that relying on the user providing the format is a good option
> * It's awesome that we now have a way to overwrite properties at runtime with the .yaml file

Agreed on using the same DataProcTemplate class, we can that to pass though to pass in arguments to the specific templates. 

Regarding properties, we could have a `--propertiesFile` argument for an actual template.properties file, which is picked up at runtime, and then remaining cli args are passed on to the specific templates (if they want them), so the general template can still have the `--config` flag. Or, we could have the config yaml file incorporate the properties. I prefer the `--propertiesFile` flag as I think it's easier to document if we keep them seperate.

Every template could also have a `--config` flag for every job, but they load their own config pojos, then they can use their own jakarta validation on pojos rather than checking properties. We could do that gradually. Regardless it will still be useful override properties at runtime, for project id etc.

bin/start.sh
```diff
@@ -97,9 +102,10 @@ case ${JOB_TYPE} in
     --jars=${JAR},${GCS_STAGING_BUCKET}/${JAR_FILE} \
     --labels job_type=dataproc_template \
     --deps-bucket=${GCS_STAGING_BUCKET} \
+    --files=${GCS_STAGING_BUCKET}/log4j-spark-driver-template.properties
     $SPARK_ARGS \
     --class com.google.cloud.dataproc.templates.main.DataProcTemplate \
-    -- ${TEMPLATE_NAME} $ARGS
+    -- --template ${TEMPLATE_NAME} $ARGS
```
tims, line 108: Template is now provided as a `--template NAME` option.

src/main/java/com/google/cloud/dataproc/templates/BaseTemplate.java
```diff
@@ -23,23 +23,14 @@
...
-    }
+    WORDCOUNT,
+    GENERAL,
+    HIVETOGCS,
+    PUBSUBTOBQ,
+    SPANNERTOGCS,
+    HIVETOBIGQUERY,
+    S3TOBIGQUERY,
+    GCSTOBIGQUERY;
   }
```
tims, line 34: Apologies for squeezing this in, but this change does not change the behaviour at all.

src/main/java/com/google/cloud/dataproc/templates/main/DataProcTemplate.java
```diff
@@ -15,86 +15,116 @@
...
-    if (args.length < 1) {
-      throw new IllegalArgumentException(
-          "Minimum 1 arguments is required: " + "<template-name> Ex wordcount");
+  /**
+   * Parse command line arguments
+   *
+   * @param args command line arguments
+   * @return parsed arguments
+   */
+  public static CommandLine parseArguments(String... args) {
```
tims, line 54: Template name and extra properties are parsed here, any other arguments are considered remainders and passed on to the template factory.

src/main/java/com/google/cloud/dataproc/templates/main/DataProcTemplate.java
```diff
@@ -15,86 +15,116 @@
...
+
+  public static void main(String[] args) {
+    CommandLine cmd = parseArguments(args);
+    String templateNameString = cmd.getOptionValue(TEMPLATE_NAME);
+    Properties properties = cmd.getOptionProperties("prop");
+    String[] remainingArgs = cmd.getArgs();
+    LOGGER.info("Template name: {}", templateNameString);
+    LOGGER.info("Properties: {}", properties);
+    LOGGER.info("Remaining args: {}", (Object) remainingArgs);
+    PropertyUtil.registerProperties(properties);
```
tims, line 89: Override the default properties that get loaded from the classpath file.

src/main/java/com/google/cloud/dataproc/templates/main/DataProcTemplate.java
```diff
@@ -15,86 +15,116 @@
...
+      ImmutableMap.<TemplateName, Function<String[], BaseTemplate>>builder()
+          .put(TemplateName.WORDCOUNT, (args) -> new WordCount())
+          .put(TemplateName.HIVETOGCS, (args) -> new HiveToGCS())
+          .put(TemplateName.HIVETOBIGQUERY, (args) -> new HiveToBigQuery())
+          .put(TemplateName.PUBSUBTOBQ, (args) -> new PubSubToBQ())
+          .put(TemplateName.GCSTOBIGQUERY, (args) -> new GCStoBigquery())
+          .put(TemplateName.S3TOBIGQUERY, (args) -> new S3ToBigQuery())
+          .put(TemplateName.SPANNERTOGCS, (args) -> new SpannerToGCS())
+          .put(TemplateName.GENERAL, GeneralTemplate::of)
+          .build();
```
tims, line 111: Using a map of template names to factories instead of a big switch statement. This also asserts that every template actually implements the BaseTemplate interface, which we didn't before.

src/main/java/com/google/cloud/dataproc/templates/main/DataProcTemplate.java
```diff
@@ -15,86 +15,116 @@
...
-        throw new IllegalArgumentException("Unexpected template name :" + template);
+    BaseTemplate template;
+    if (templateFactories.containsKey(templateName)) {
+      template = templateFactories.get(templateName).apply(args);
+    } else {
+      throw new IllegalArgumentException(
+          String.format("Unexpected template name: %s", templateName));
     }
-
+    template.runTemplate();
```
tims, line 127: The templates still have to start their own spark session because that's too big a change for this PR, but this would be the perfect place to start, configure and then pass in the spark session to the runTemplate method, to improve consistency and readability of the templates.

src/test/java/com/google/cloud/dataproc/templates/main/DataProcTemplateTest.java
```diff
@@ -31,32 +35,66 @@
   @ParameterizedTest
   @MethodSource("stringValidInputArgs")
   void testRunSparkJobWithValidInputArgs(List<String> args) {
-    DataProcTemplate.main((String[]) args.toArray());
+    DataProcTemplate.main(args.toArray(new String[0]));
```
tims, line 38: Fixed failing tests from incorrect casting.

src/test/resources/log4j-spark-driver-template.properties
```diff
@@ -0,0 +1,21 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+#
+
+# Set everything to be logged to the console
+log4j.rootCategory=INFO, console
+log4j.appender.console=org.apache.log4j.ConsoleAppender
+log4j.appender.console.layout=org.apache.log4j.PatternLayout
+log4j.appender.console.layout.ConversionPattern=%d{yy/MM/dd HH:mm:ss} %p %c{1}: %m%n
```
tims, line 21: This log4j config is moved to src/test/resources because it was not being found on the class path anyway, so now the start.sh script passes it to the jobs as a file.

tims:
Closing and will reopen a clean PR with the changes we've discussed offline.

PR #32, opened 2021-12-14
Refactor start.sh to allow passing properties to templates

tims:
Can now pass --template and --templateProperty arguments. 

Start.sh passes any args, directly to the spark submit command. So spark flags can be passed first, then template options such as a new `--template NAME` flag and `--templateProperty key=value` flag can be passed after a `--` separator. 

Variables that specific to the bash script, rather than the spark or template options are no longer passed as arguments (and parsed by bash), they are instead passed an environment variables. So can be easily set once with `export VAR=FOO` bash statements elsewhere, or passed inline like this example:
```
GCP_PROJECT=<gcp-project-id> \
REGION=<region>  \
SUBNET=<subnet>   \
GCS_STAGING_LOCATION=<gcs-staging-bucket-folder> \
HISTORY_SERVER_CLUSTER=<history-server> \
bin/start.sh \
-- --template GCSTOBIGQUERY
--templateProperty project.id=<gcp-project-id> \
--templateProperty gcs.bigquery.input.location=<gcs path> \
--templateProperty gcs.bigquery.input.format=<csv|parquet|avro> \
--templateProperty gcs.bigquery.output.dataset=<datasetId> \
--templateProperty gcs.bigquery.output.table=<tableName> \
--templateProperty gcs.bigquery.temp.bucket.name=<bigquery temp bucket name>
```


franklinWhaite (commented):
When running with JOB_TYPE=CLUSTER the start.sh may pick up env variables such as `deps-bucket`,` histroy-server-cluser `and subnet and use these for the dataproc cluster job. This will cause an error wince the `gcloud dataproc jobs submit spark` is not expecting these arguments

franklinWhaite:
Could you please update the README in the root directory

tims:
Updated the root readme, and tweaked the start.sh to skip options not needed in cluster mode.

franklinWhaite (approved):
LGTM

tims:
Fixed readme typo, and added a JOB_TYPE env var for setting CLUSTER or SERVERLESS mode (defaults to SERVERLESS).

PR #33, opened 2021-12-15
removing dependencies on log4j versions <2.16.0, adding log4j version…

franklinWhaite:


tims (approved):
Confirmed.

Noting:

This excludes the old 1.x log4j (that isn't vulnerable anyway) from the aws hadoop dependency. 

And this successfully explicitly includes the 2.16 version, which will override any transitive includes to protect against transitive dependencies on vulnerable versions in the future (though there are non at the moment).

PR #35, opened 2021-12-17
Add general template

tims:
General template #29 

can run via 
```
bin/start.sh \
--files=gs://bucket/path/config.yaml \
-- \
--template GENERAL \
--config config.yaml
```

with config.yaml eg:
```
input:
  shakespeare:
    format: bigquery
    options:
      table: "bigquery-public-data:samples.shakespeare"
query:
  wordcount:
    sql: "SELECT word, sum(word_count) cnt FROM shakespeare GROUP by word ORDER BY cnt DESC"
output:
  wordcount:
    format: csv
    options:
      header: true
    path: gs://bucket/output/wordcount/
    mode: Overwrite
```


franklinWhaite:
I believe that the general template is missing in the `com.google.cloud.dataproc.templates.main.DataProcTemplate`.

When running `bin/start.sh --files="gs://bucket/path/config.yaml"  -- --template GENERAL --config config.yaml` I get `Exception in thread "main" java.lang.IllegalArgumentException: Unexpected template name: GENERAL`

tims:
> 

Have added general to template factory now.

franklinWhaite (approved):
LGTM

PR #70, opened 2022-02-21
Jdbc to big query template

snrssc:
Adding new template JDBC to BigQuery


src/main/java/com/google/cloud/dataproc/templates/util/TemplateConstants.java
```diff
@@ -65,6 +65,18 @@
...
 
+  /** Property values for JDBCToBQ */
+  String JDBC_TO_BQ_BIGQUERY_LOCATION = "jdbctobq.bigquery.location";
+
+  String JDBC_TO_BQ_JDBC_URL = "jdbctobq.jdbc.url";
+  String JDBC_TO_BQ_JDBC_DRIVER_CLASS_NAME = "jdbctobq.jdbc.driver.class.name";
+  String JDBC_TO_BQ_INPUT_TABLE_PROP = "jdbctobq.input.table";
+  String JDBC_TO_BQ_INPUT_TABLE_DATABASE_PROP = "jdbctobq.input.db";
+  String JDBC_TO_BQ_APPEND_MODE = "jdbctobq.append.mode";
+  String JDBC_TO_BQ_PARTITION_COL = "jdbctobq.partition.col";
```
franklinWhaite, line 76: It seems like this constant is not used, guess it can be deleted?

src/main/resources/template.properties
```diff
@@ -32,6 +32,22 @@ hivetobq.partition.col=
...
+jdbctobq.bigquery.location=
+jdbctobq.jdbc.url=jdbc:mysql://host:port/
+jdbctobq.jdbc.driver.class.name=com.mysql.jdbc.Driver
+#Customize any properties supported by jdbc. Value to be in JSON format
+jdbctobq.jdbc.properties={"user":"usernamehere","password":"passwordhere"}
+##jdbc:bigquery://https://www.googleapis.com/bigquery/v2:443;ProjectId=nsadineni;OAuthType=1;
+jdbctobq.input.table=
+jdbctobq.input.db=
+#Write mode to use while writing output to BQ. Supported values are - Append/Overwrite/ErrorIfExists/Ignore
+jdbctobq.append.mode=ErrorIfExists
```
franklinWhaite, line 46: is _append.mode_ the best name for this variable? Should it not be something like _write.mode_

franklinWhaite (approved):
LGTM. I found some minor naming issues, but I am ok merging despite those issues. Feel free to fix now and merge or fix in next commit.

src/main/java/com/google/cloud/dataproc/templates/util/TemplateConstants.java
snrssc: Removed the unused variables

src/main/resources/template.properties
snrssc: Thanks Frank. I have modified the variable name to write.mode

Issue #45, opened 2022-01-12
New template for "PubSub to GCS"


ajayydv:
PubSub to GCS:

-  avro 
- parquet
- csv

PR #88, opened 2022-03-29
Pubsub to gcs

jirilmongeorge-google:
Template Name: **PubSub to GCS**
Template Task Number: https://github.com/GoogleCloudPlatform/dataproc-templates/issues/45
Supported output formats: AVRO and JSON

Sample execution output available in bucket: /pubsubtogcs_dev/

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Google is providing this collection of pre-implemented Dataproc templates as a r
 * [JDBCToBigQuery](src/main/java/com/google/cloud/dataproc/templates/jdbc/README.md)
 * [JDBCToGCS](src/main/java/com/google/cloud/dataproc/templates/jdbc/README.md)
 * [WordCount](src/main/java/com/google/cloud/dataproc/templates/word/WordCount.java)
-* [PubSubToGCS](src/main/java/com/google/cloud/dataproc/templates/pubsub/README.md)
+* [PubSubToGCS](src/main/java/com/google/cloud/dataproc/templates/pubsub/README.md#2-pubsub-to-gcs)
 * [GeneralTemplate](src/main/java/com/google/cloud/dataproc/templates/general/README.md)
 
 
@@ -150,7 +150,7 @@ Google is providing this collection of pre-implemented Dataproc templates as a r
        bin/start.sh -- --template BIGQUERYTOGCS
        ```
    1. #### Executing PubSub to GCS template.
-      Detailed instructions at [README.md](src/main/java/com/google/cloud/dataproc/templates/pubsub/README.md)
+      Detailed instructions at [README.md](src/main/java/com/google/cloud/dataproc/templates/pubsub/README.md#2-pubsub-to-gcs)
       ```
       bin/start.sh -- --template PUBSUBTOGCS
       ```
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ Google is providing this collection of pre-implemented Dataproc templates as a r
 * [JDBCToBigQuery](src/main/java/com/google/cloud/dataproc/templates/jdbc/README.md)
 * [JDBCToGCS](src/main/java/com/google/cloud/dataproc/templates/jdbc/README.md)
 * [WordCount](src/main/java/com/google/cloud/dataproc/templates/word/WordCount.java)
+* [PubSubToGCS](src/main/java/com/google/cloud/dataproc/templates/pubsub/README.md)
 * [GeneralTemplate](src/main/java/com/google/cloud/dataproc/templates/general/README.md)
 
 
@@ -148,6 +149,11 @@ Google is providing this collection of pre-implemented Dataproc templates as a r
        ```
        bin/start.sh -- --template BIGQUERYTOGCS
        ```
+   1. #### Executing PubSub to GCS template.
+      Detailed instructions at [README.md](src/main/java/com/google/cloud/dataproc/templates/pubsub/README.md)
+      ```
+      bin/start.sh -- --template PUBSUBTOGCS
+      ```
 
     1. #### Executing General template.
        Detailed instructions at [README.md](src/main/java/com/google/cloud/dataproc/templates/general/README.md)
--- a/pom.xml
+++ b/pom.xml
@@ -229,6 +229,19 @@
       <artifactId>hadoop-common</artifactId>
       <version>3.2.2</version>
     </dependency>
+
+    <!-- for pubsub to gcs -->
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-storage</artifactId>
+      <version>2.2.2</version>
+    </dependency>
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
+      <version>20211205</version>
+    </dependency>
+
   </dependencies>
 
   <build>
--- a/src/main/java/com/google/cloud/dataproc/templates/pubsub/PubSubToGCS.java
+++ b/src/main/java/com/google/cloud/dataproc/templates/pubsub/PubSubToGCS.java
@@ -16,6 +16,7 @@
 package com.google.cloud.dataproc.templates.pubsub;
 
 import static com.google.cloud.dataproc.templates.util.TemplateConstants.*;
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 import com.google.cloud.dataproc.templates.BaseTemplate;
 import com.google.cloud.storage.Blob;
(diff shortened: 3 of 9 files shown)


src/main/java/com/google/cloud/dataproc/templates/pubsub/README.md
```diff
@@ -35,3 +35,52 @@ pubsub.bq.output.table=<bq output table>
 ## Number of records to be written per message to BigQuery
 pubsub.bq.batch.size=1000
 ```
+
+## 2. Pub/Sub To GCS
+
```
shashank-google, line 40: We need to provide details about assumption and requirements for pubsub message payload format etc if any.
It seems like we are requiring customer to have JSON as payload but that is nowhere mentioned in readme.
Also how those column names are mapped to GCS / Avro columns and what happens if there is array / map / collections in the json payload.

src/main/java/com/google/cloud/dataproc/templates/util/TemplateConstants.java
```diff
@@ -193,4 +193,27 @@
...
+  String PUBSUB_GCS_TOTAL_RECEIVERS_PROP = "pubsubtogcs.total.receivers";
+  // Project that contains the GCS output
+  String PUBSUB_GCS_OUTPUT_PROJECT_ID_PROP = "pubsubtogcs.gcs.output.project.id";
+  // GCS bucket URL
+  String PUBSUB_GCS_BUCKET_NAME = "pubsubtogcs.gcs.bucket.name";
+  // Number of records to be written per message to GCS
+  String PUBSUB_GCS_BATCH_SIZE_PROP = "pubsubtogcs.batch.size";
+  String PUBSUB_GCS_BUCKET_OUTPUT_PATH = "output/";
+  String PUBSUB_GCS_OUTPUT_DATA_FORMAT = "pubsubtogcs.gcs.output.data.format";
+  String[] PUBSUB_GCS_OUTPUT_FORMATS_ARRAY = {"AVRO", "JSON"};
```
shashank-google, line 216: We do not need this static array declaration. 
It might be good to use straight string values from.

src/main/java/com/google/cloud/dataproc/templates/pubsub/PubSubToGCS.java
```diff
@@ -0,0 +1,227 @@
...
+        PUBSUB_GCS_OUTPUT_PROJECT_ID_PROP,
+        outputProjectID,
+        PUBSUB_GCS_BUCKET_NAME,
+        gcsBucketName,
+        PUBSUB_GCS_BATCH_SIZE_PROP,
+        batchSize,
+        PUBSUB_GCS_OUTPUT_DATA_FORMAT,
+        outputDataFormat);
+
+    try {
```
shashank-google, line 122: lets remove try / catch. We do not want it going forward in the templates.

pom.xml
```diff
@@ -229,6 +229,19 @@
       <artifactId>hadoop-common</artifactId>
       <version>3.2.2</version>
     </dependency>
+
+    <!-- for pubsub to gcs -->
```
shashank-google, line 233: Are we sure we need new dependencies.
There are plenty of existing templates with PubSub, GCS and JSON so Ideally we should not be needing it.

README.md
```diff
@@ -148,6 +149,11 @@ Google is providing this collection of pre-implemented Dataproc templates as a r
        ```
        bin/start.sh -- --template BIGQUERYTOGCS
        ```
+   1. #### Executing PubSub to GCS template.
```
shashank-google, line 152: I would prefer to remove this part of readme as it does not adds much value.
Separate readme file is already linked on the top.

src/main/java/com/google/cloud/dataproc/templates/pubsub/PubSubToGCS.java
```diff
@@ -0,0 +1,227 @@
...
+      Bucket bucket = storage.get(gcsBucketName);
+      writeToGCS(stream, outputProjectID, gcsBucketName, batchSize, outputDataFormat, bucket);
+
+      jsc.start();
+      jsc.awaitTerminationOrTimeout(timeoutMs);
+
+      LOGGER.info("PubSubToGCS job completed.");
+      jsc.stop();
+
+    } catch (Throwable th) {
```
shashank-google, line 153: lets remove try / catch it is not needed.

src/main/java/com/google/cloud/dataproc/templates/pubsub/PubSubToGCS.java
```diff
@@ -0,0 +1,227 @@
...
+      LOGGER.error("Exception in PubSubToGCS", th);
+      if (Objects.nonNull(jsc)) {
+        jsc.stop();
+      }
+    }
+  }
+
+  public static void writeToGCS(
```
shashank-google, line 161: too long method needs to be sub divided into more methods.
Also I feel whole of this method is not needed. 
Refer: https://github.com/GoogleCloudPlatform/dataproc-templates/blob/main/src/main/java/com/google/cloud/dataproc/templates/databases/SpannerToGCS.java#L74

src/main/java/com/google/cloud/dataproc/templates/pubsub/PubSubToGCS.java
```diff
@@ -0,0 +1,227 @@
...
+        Integer.parseInt(getProperties().getProperty(PUBSUB_GCS_STREAMING_DURATION_SECONDS_PROP));
+    totalReceivers = Integer.parseInt(getProperties().getProperty(PUBSUB_GCS_TOTAL_RECEIVERS_PROP));
+    outputProjectID = getProperties().getProperty(PUBSUB_GCS_OUTPUT_PROJECT_ID_PROP);
+    gcsBucketName = getProperties().getProperty(PUBSUB_GCS_BUCKET_NAME);
+    batchSize = Integer.parseInt(getProperties().getProperty(PUBSUB_GCS_BATCH_SIZE_PROP));
+    outputDataFormat = getProperties().getProperty(PUBSUB_GCS_OUTPUT_DATA_FORMAT);
+  }
+
+  @Override
+  public void runTemplate() {
```
shashank-google, line 72: too long of method, can we please try to maintain it <20 line of code / method.

src/main/java/com/google/cloud/dataproc/templates/pubsub/PubSubToGCS.java
```diff
@@ -0,0 +1,227 @@
...
+                StorageLevel.MEMORY_AND_DISK_SER());
+        if (stream == null) {
+          stream = pubSubReciever;
+        } else {
+          stream = stream.union(pubSubReciever);
+        }
+      }
+
+      LOGGER.info("Writing data to output GCS Bucket: {}", gcsBucketName);
+      Storage storage = StorageOptions.getDefaultInstance().getService();
```
shashank-google, line 143: Why is the code getting the storage bucket. 
Spark can do it automatically.
Refer: https://github.com/GoogleCloudPlatform/dataproc-templates/blob/main/src/main/java/com/google/cloud/dataproc/templates/databases/SpannerToGCS.java#L74

src/main/java/com/google/cloud/dataproc/templates/pubsub/PubSubToGCS.java
```diff
@@ -0,0 +1,227 @@
...
+      Bucket bucket = storage.get(gcsBucketName);
+      writeToGCS(stream, outputProjectID, gcsBucketName, batchSize, outputDataFormat, bucket);
```
shashank-google, line 145: This method feel unnecessary.

shashank-google (changes requested):

src/main/java/com/google/cloud/dataproc/templates/pubsub/PubSubToGCS.java
jirilmongeorge-google: @shashank-google , The SpannerToGCS uses Dataframe approach. I was following the same approach used in PubsubToBigQuery template. I could not find any code samples for reading PubSub data using Spark Dataframes. Please let me know if we have any references for it.

src/main/java/com/google/cloud/dataproc/templates/pubsub/README.md
jirilmongeorge-google: Hi @shashank-google , this solution supports AVRO and JSOn data formats. It does not require customer to follow any particular schema.  If AVRO data is passed through pubsub topic then it will be stored as .avro files in GCS.If the  pubsub topic had JSON data passed through it, we can mention JSON as the file format on solution and it will produce .json files in GCS. Hope this clarifies.

shashank-google (changes requested):

shashank-google:
@jirilmongeorge-google bumping up.
Can you please update the pull request to address the comments.

PR #99, opened 2022-04-20
Python Templates

### LINKED ISSUE #86: New Template: Python template for GCS To BigQuery

### LINKED ISSUE #87: New Template: Python template for BigQuery To GCS

NiloFreitas:
Implement all the structure necessary for running PySpark templates.
It segregates Python and Java implementation into separate folders.
It implements a specific bin/start.sh to run PySpark on Dataproc Serverless.

Argument parsing is currently handled differently than in the Java implementation:
  - There are no global defauls for properties. Each template defines each own arguments. Doing it this way reduces clutter and allows for better help messages.
  - The properties are passed directly, without the --templateProperty key. For example, `--templatePropery=my.property=value` becomes `--my.property=value`.
 
It implements 2 templates in PySpark:
  - GCSToBigQuery
  - BigQueryToGCS
  
Closes #86
Closes #87


ppaglilla:
One thing to note is that, because the project was reorganized into the `java` and `python` directories, the Jenkins pipelines we have in place will need to be updated. We simply need to accommodate for the fact that the Java project is no longer located at the root of the repository.

To that end, we created the `build-job-python` and `dataproc-teleport-integration-tests-python` pipelines. Those are clones of the `build-job` and `dataproc-teleport-integration-tests` pipelines, but they run against the `python` branch with the necessary changes applied. The new pipelines do not run Python unit tests or Python integration tests for the moment.

Before merging this PR, the original pipelines should be updated.

python/dataproc_templates/gcs/README.md
```diff
@@ -0,0 +1,21 @@
...
+```
+export GCP_PROJECT=<project_id> 
+export SUBNET=<region> 
+export JARS="gs://spark-lib/bigquery/spark-bigquery-latest_2.12.jar" 
+export GCS_STAGING_LOCATION=<gcs-staging-bucket-folder> 
+export REGION=<region>
+
+./bin/start.sh \
+-- --template=GCSTOBIGQUERY \
+    --gcs.bigquery.input.format="<csv|parquet|avro>" \
```
franklinWhaite, line 14: why is json not supported?

franklinWhaite, line 14: why is json not supported?

python/README.md
```diff
@@ -0,0 +1,69 @@
...
+                  --my.property="<value>" \
+                  --my.other.property="<value>"
+                  (etc...)
+```
+To see template's specific parameters, refer to each template's README.  
+Refer to this [documentation](https://cloud.google.com/dataproc-serverless/docs/concepts/properties) to see the available spark properties.
+
+It is also possible to submit the jobs using gcloud CLI, after building the package.
+
+**gcloud CLI usage syntax**:
```
franklinWhaite, line 49: this is cool!

franklinWhaite, line 49: this is cool!

python/bin/dataproc_template_functions.sh
```diff
@@ -0,0 +1,67 @@
...
+
+#Mandatory vs optional  specify
+Help() {
+  # Display Help
+  help_text=$(cat << EndOfMessage
+    Usage:
+
+    # Environment variables
+    export GCP_PROJECT=projectId
+    export REGION=us-west1
```
franklinWhaite, line 40: I believe we should not have actual values in the example

franklinWhaite, line 40: I believe we should not have actual values in the example

python/bin/start.sh
```diff
@@ -0,0 +1,79 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#Initialize functions and Constants
+BIN_DIR="$(dirname "$BASH_SOURCE")"
+PROJECT_ROOT_DIR=${BIN_DIR}/..
+
+PACKAGE_EGG_FILE=dist/dataproc_templates-0.0.1-py3.8.egg
```
franklinWhaite, line 22: I am not very string on this one, but ...
- On one side since this name includes python version number, it will break for users with slightly different versions (e.g. Python 3.9)

- On the other side, since we specified Python 3.8 in requirements should we worry too much about supporting other versions?

I understand this is the default name when creating the Python egg, but name could easily be modified after creation.

franklinWhaite, line 22: I am not very string on this one, but ...
- On one side since this name includes python version number, it will break for users with slightly different versions (e.g. Python 3.9)

- On the other side, since we specified Python 3.8 in requirements should we worry too much about supporting other versions?

I understand this is the default name when creating the Python egg, but name could easily be modified after creation.

python/dataproc_templates/bigquery/README.md
```diff
@@ -0,0 +1,20 @@
...
+export SUBNET=<region>
+export JARS="gs://spark-lib/bigquery/spark-bigquery-latest_2.12.jar"
+export GCS_STAGING_LOCATION=<gcs-staging-bucket-folder>
+export REGION=<region>
+
+./bin/start.sh \
+-- --template=BIGQUERYTOGCS \
+	--bigquery.gcs.input.table=<projectId:datasetId.tableName> \
+	--bigquery.gcs.output.format=<csv|parquet|avro|json> \
+	--bigquery.gcs.output.mode=<overwrite|append> \
```
franklinWhaite, line 16: Usually spark has 4 save modes: overwrite, append, ignore, errorifexists ... is there any reason for the exclusion of the latter 2 here? are they not supported when writing to GCS?

franklinWhaite, line 16: Usually spark has 4 save modes: overwrite, append, ignore, errorifexists ... is there any reason for the exclusion of the latter 2 here? are they not supported when writing to GCS?

python/dataproc_templates/bigquery/bigquery_to_gcs.py
```diff
@@ -0,0 +1,120 @@
...
+
+        logger.info(
+            "Starting Bigquery to GCS spark job with parameters:\n"
+            f"{pprint.pformat(args)}"
+        )
+
+        # Read
+        input_data = spark.read \
+            .format("bigquery") \
+            .option("table", input_table) \
```
franklinWhaite, line 95: nit. these regular expressions could be constants

franklinWhaite, line 95: nit. these regular expressions could be constants

python/dataproc_templates/gcs/gcs_to_bigquery.py
```diff
@@ -0,0 +1,115 @@
...
+                .option(constants.GCS_BQ_CSV_HEADER, True) \
+                .option(constants.GCS_BQ_CSV_INFER_SCHEMA, True) \
+                .load(input_file_location)
+
+        # Write
+        input_data.write \
+            .format(constants.GCS_BQ_OUTPUT_FORMAT) \
+            .option(constants.GCS_BQ_OUTPUT, big_query_dataset + "." + big_query_table) \
+            .option(constants.GCS_BQ_TEMP_BUCKET, bq_temp_bucket) \
+            .mode("append") \
```
franklinWhaite, line 114: This should e a template property. User should be able to specify the save mode (i.e. overwrite, append, errorifexists, ignore)

franklinWhaite, line 114: This should e a template property. User should be able to specify the save mode (i.e. overwrite, append, errorifexists, ignore)

python/main.py
```diff
@@ -0,0 +1,125 @@
...
+    )
+
+    LOGGER.info('Running template %s', template_name)
+
+    template_instance: BaseTemplate = template_impl.build()
+
+    try:
+        args: Dict[str, Any] = template_instance.parse_args()
+        spark: SparkSession = create_spark_session(template_name=template_name)
+        template_instance.run(spark=spark, args=args)
```
franklinWhaite, line 111: This is cool, having the main create the spark session!

franklinWhaite, line 111: This is cool, having the main create the spark session!

python/dataproc_templates/gcs/README.md
ppaglilla: I see no reason why it should not be supported, so I went ahead and added it.

One thing to note is that JSON is not supported in the Java version of this template. Should we consider adding it as well?

python/bin/start.sh
ppaglilla: Good catch! I had thought about the package version potentially being a problem, not realizing we had the Python version there as well. Ideally, we'd want the start script to work with any package/python version. 

To that end, I customized the build command to allow setting a custom name for the build artifact. Now, we don't rely on the package having a particular name. See [f2fe3f1](https://github.com/GoogleCloudPlatform/dataproc-templates/pull/99/commits/f2fe3f1237faf10077e8e7c9efb89a783bfb74e5).

python/dataproc_templates/bigquery/README.md
ppaglilla: They are supported, yes! I added them to the template.

python/dataproc_templates/gcs/gcs_to_bigquery.py
ppaglilla: Added! Now the save mode is an argument and it can be set to any of the 4 modes available. I set it to default to `append`, but we could make it mandatory. Thoughts?

One thing to note is that the Java version of this template also hard-codes `append`. Should we consider making it a parameter as well?

python/bin/dataproc_template_functions.sh
ppaglilla: Do you mean keeping the values empty for the example?

python/dataproc_templates/gcs/README.md
franklinWhaite: great!
Per the java template, if that is the case, I would recommend creating a separate issue for that.

python/bin/dataproc_template_functions.sh
franklinWhaite: (this is also nitpicking)
But for example in `export GCP_PROJECT=projectId` no real project id is provided. 
We could have `export REGION=region`

python/dataproc_templates/bigquery/README.md
franklinWhaite: great!

per the java template, I would recommend creating a separate issue for that!

python/dataproc_templates/gcs/gcs_to_bigquery.py
franklinWhaite: Great!

imo it does not matter that much if it is mandatory or optional as long as the default value is clearly specified in the readme

per the java template, I would recommend creating a separate issue.

python/bin/dataproc_template_functions.sh
NiloFreitas: Changed the example to "region" :)

python/dataproc_templates/bigquery/bigquery_to_gcs.py
NiloFreitas: Refactored constants, and fixed this in [#de4731c](https://github.com/GoogleCloudPlatform/dataproc-templates/commit/de4731c03edbd7122e2eb1ed3329747dd7c117c9)

python/dataproc_templates/bigquery/README.md
```diff
@@ -0,0 +1,20 @@
+## BigQuery To GCS 
+
+General Execution:
+
+```
+export GCP_PROJECT=<project_id>
+export SUBNET=<region>
+export JARS="gs://spark-lib/bigquery/spark-bigquery-latest_2.12.jar"
```
shashank-google, line 8: Are we sure we need this jar with python ?

python/dataproc_templates/bigquery/bigquery_to_gcs.py
```diff
@@ -0,0 +1,119 @@
...
+        )
+
+        known_args: argparse.Namespace
+        known_args, _ = parser.parse_known_args(args)
+
+        return vars(known_args)
+
+    def run(self, spark: SparkSession, args: Dict[str, Any]) -> None:
+
+        log4jLogger = spark.sparkContext._jvm.org.apache.log4j  # pylint: disable=protected-access,invalid-name
```
shashank-google, line 85: is pylint comment supposed to be here ?

python/dataproc_templates/gcs/README.md
```diff
@@ -0,0 +1,22 @@
+## GCS To BigQuery
+
+General Execution:
+
+```
+export GCP_PROJECT=<project_id> 
+export SUBNET=<region> 
+export JARS="gs://spark-lib/bigquery/spark-bigquery-latest_2.12.jar"
```
shashank-google, line 8: jar ?

python/dataproc_templates/gcs/README.md
```diff
@@ -0,0 +1,22 @@
+## GCS To BigQuery
+
+General Execution:
+
+```
+export GCP_PROJECT=<project_id> 
+export SUBNET=<region>
```
shashank-google, line 7: do we really need subnet here ?
Can we try and test without subnet ?

python/dataproc_templates/bigquery/README.md
```diff
@@ -0,0 +1,20 @@
+## BigQuery To GCS 
+
+General Execution:
+
+```
+export GCP_PROJECT=<project_id>
+export SUBNET=<region>
```
shashank-google, line 7: subnet ?
lets try without subnet, if it works.

shashank-google (commented):
minor comments

python/dataproc_templates/gcs/README.md
surjits254: I tested this without subnet. "gcloud beta dataproc batches submit pyspark" requires a subnet with private google access enabled. It uses default subnet if not explicitly provided. Our sandbox environment default subnet does not have private google access enabled.

python/dataproc_templates/bigquery/README.md
surjits254: responded on previous comment

python/dataproc_templates/bigquery/bigquery_to_gcs.py
ppaglilla: It is. By default, pylint flags accessing protected fields in an object (anything starting with an underscore) as an error and warns you about it. However; getting the Spark logger from Python requires accessing `SparkContext._jvm`, which triggers the warning.

With that comment, we acknowledge we're accessing a protected field and disable the `protected-access` check only for that line. That way; accessing `_jvm` does not trigger a linter warning, but we get to keep that check for the rest of the project.

We're also disabling `invalid-name`, but that avoid that by just renaming the variable to use snake case. Will do that later!

python/dataproc_templates/gcs/README.md
ppaglilla: We can move the subnet to the optional options on the documentation. I'll be pushing some updates to the docs later, so I'll update it then.

python/dataproc_templates/bigquery/README.md
surjits254: yes, pyspark requires connector to be manually specified

python/dataproc_templates/gcs/README.md
surjits254: responded on previous comment

python/dataproc_templates/bigquery/README.md
ppaglilla: To elaborate here a bit, the Java implementation ships with many JARs when it's packaged into a JAR itself. For example, the Spark BigQuery connector is shipped right with the templates instead of being specified at job submission time.

For Python, shipping JARs with the code it not an option, they need to be passed at job submission.

We could document clearly the required JARs for each template (and maybe provide recommended URIs to them if available).

PR #120, opened 2022-05-10
Jdbc read partitioning

shashank-google:


java/src/main/java/com/google/cloud/dataproc/templates/jdbc/JDBCToGCS.java
```diff
@@ -65,17 +80,40 @@ public void runTemplate() {
               + "in resources/conf/template.properties file or at runtime. Refer to jdbc/README.md for more instructions.");
     }
 
+    if (StringUtils.isNotBlank(concatedPartitionProps)
```
snrssc, line 83: This logic makes parallel read as mandatory and job fails if partitioning is not specified.

java/src/main/java/com/google/cloud/dataproc/templates/jdbc/JDBCToGCS.java
```diff
@@ -65,17 +80,40 @@ public void runTemplate() {
...
+            + "5. {}:{}"
+            + "6. {}:{}"
+            + "7. {}:{}",
         JDBC_TO_GCS_OUTPUT_LOCATION,
-        GCSOutputLocation,
+        gcsOutputLocation,
         JDBC_TO_GCS_WRITE_MODE,
-        GCSWriteMode,
+        gcsWriteMode,
         JDBC_TO_GCS_JDBC_URL,
```
snrssc, line 107: Printing JDBC URL in logs printing password also in logs. May be we can remove password and print or simply ignore printing jdbc url for now. It's okay to handle this in a new issue too.

snrssc (commented):
Except the logic making partition read mandatory, everything else looks good.

java/src/main/java/com/google/cloud/dataproc/templates/jdbc/JDBCToGCS.java
shashank-google: No. that is not correct.
It is a multiline condition If (concatedPartitionProps != blank && (jdbcSQLPartitionColumn != blank || ...)) then trow exception.

java/src/main/java/com/google/cloud/dataproc/templates/jdbc/JDBCToGCS.java
shashank-google: Good point thanks. Removed it.

PR #121, opened 2022-05-12
removing support for migration of several entities in a single run

franklinWhaite:
Originally the template could migrate several tables in a single batch.
This PR simplifies the template by limiting it's functionality to a single table per run.

This is because handling table specific arguments while migrating several tables in a single batch was adding unnecessary complexity.


ppaglilla (approved):
Looks good to me. It just simplifies the logic that was already in place.

Issue #108, opened 2022-05-03
[New Template] [Python] GCS Avro / Parquet etc to Cloud Spanner


shashank-google:
Import data into cloud spanner single table from GCS avro and parquet formats.

shashank-google:
Spanner templates currently require Spanner JDBC Dialect and therefore it is not feasible for now with Python.
Out of scope.

PR #122, opened 2022-05-13
PySpark - GCS to Spanner

hec10r:
Issue #108


python/dataproc_templates/gcs/README.md
```diff
@@ -63,3 +63,76 @@ export JARS="gs://spark-lib/bigquery/spark-bigquery-latest_2.12.jar"
...
+## GCS To Spanner
+
+Template for reading files from Google Cloud Storage and writing them to a
+Spanner table. It supports reading Parquet, Avro and ORC formats.
+
+## Arguments
+
+* `gcs.spanner.input.location`: GCS location of the input file
+(format: `gs://BUCKET/...`)
+* `gcs.spanner.input.format`: Input file format (one of: avro, parquet, orc)
```
ppaglilla, line 76: Can we support CSV and JSON as well?

I know the Java implementation only support Parquet, Avro and ORC; but it's very common to see client use CSV and JSON as well. If we add support for them, we could also backport that support to the Java templates.

python/dataproc_templates/gcs/gcs_to_spanner.py
```diff
@@ -0,0 +1,160 @@
...
+    )
+    parser.add_argument(
+      f'--{constants.GCS_SPANNER_OUTPUT_SAVE_MODE}',
+      dest=constants.GCS_SPANNER_OUTPUT_SAVE_MODE,
+      required=False,
+      default=constants.OUTPUT_MODE_APPEND,
+      help=(
+        'Output write mode '
+        '(one of: <Overwrite|ErrorIfExists|Append|Ignore>) '
+        '(Defaults to ErrorIfExists)'
```
ppaglilla, line 90: The help and the docs in the README state the save mode defaults to `errorifexists`, but it actually defaults to `append`. Other Python templates default to `append`; so I'd suggest keeping that default, but updating the docs and help.

As a side note, I think defaulting to `errorifexists` is worth discussing. That's [Spark's built-in default](https://spark.apache.org/docs/latest/sql-data-sources-load-save-functions.html#save-modes) and we should probably respect that to be consistent with the ecosystem.

python/dataproc_templates/gcs/gcs_to_spanner.py
```diff
@@ -0,0 +1,160 @@
...
+        constants.OUTPUT_MODE_IGNORE,
+        constants.OUTPUT_MODE_ERRORIFEXISTS
+      ]
+    )
+    parser.add_argument(
+      f'--{constants.GCS_SPANNER_OUTPUT_BATCH_INSERT_SIZE}',
+      dest=constants.GCS_SPANNER_OUTPUT_BATCH_INSERT_SIZE,
+      required=False,
+      default=1000,
+      help='Spanner batch insert size '
```
ppaglilla, line 104: The default batch size should be documented in the help and the README, so that users know the default value without needing to jump into the code

python/dataproc_templates/gcs/gcs_to_spanner.py
```diff
@@ -0,0 +1,160 @@
...
+    input_data: DataFrame
+
+    input_data = (
+      spark.read
+      .format(input_file_format)
+      .load(input_file_location)
+    )
+
+    # Write
+    (
```
ppaglilla, line 149: Can we change the formatting to be consistent with the rest of the templates?

An example from GCS to BigQuery:

https://github.com/GoogleCloudPlatform/dataproc-templates/blob/f864cd97ca96f5e3da99931f323d7c9a52035094/python/dataproc_templates/gcs/gcs_to_bigquery.py#L134-L139

python/dataproc_templates/gcs/README.md
```diff
@@ -63,3 +63,76 @@ export JARS="gs://spark-lib/bigquery/spark-bigquery-latest_2.12.jar"
...
+* `gcs.spanner.output.instance`: Spanner instance name
+* `gcs.spanner.output.database`: Spanner database name
+* `gcs.spanner.output.table`: Spanner table name
+* `gcs.spanner.output.primary_key`: Primary Key of the Spanner table
```
ppaglilla, line 80: We should add the fact that the PK is only required for new tables here

python/dataproc_templates/gcs/gcs_to_spanner.py
```diff
@@ -0,0 +1,160 @@
...
+      required=True,
+      help='Spanner output table'
+    )
+    parser.add_argument(
+      f'--{constants.GCS_SPANNER_PRIMARY_KEY}',
+      dest=constants.GCS_SPANNER_PRIMARY_KEY,
+      required=False,
+      help=(
+        "Primary key of the table. "
+        "It is only required when the table doesn't exist"
```
ppaglilla, line 79: Do we know what happens when the table does not exist, but the PK is not provided?

Ideally; user's would get a sensible error message when this happens.   We should check if the JDBC driver error is good enough and, in case it's not, see if we can provide a better one.

ppaglilla (changes requested):

python/dataproc_templates/gcs/README.md
hec10r: The _problem_ is that CSV files usually require additional configuration and I wanted to keep this simple. The support may be added in a different PR.

python/dataproc_templates/gcs/gcs_to_spanner.py
hec10r: You are right. I am changing the default to `errorifexists` for consistency (since this way is defined for the Java template)

python/dataproc_templates/gcs/gcs_to_spanner.py
hec10r: Yeah, we need a linter urgently. In the meantime I will change it for consistency but I strongly prefer the way I did it.

python/dataproc_templates/gcs/README.md
hec10r: Agree, and in the Java implementation as well: https://github.com/GoogleCloudPlatform/dataproc-templates/blob/main/java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCSToSpannerConfig.java#L75

python/dataproc_templates/gcs/gcs_to_spanner.py
hec10r: Agree. This is pending

python/dataproc_templates/gcs/gcs_to_spanner.py
ppaglilla: Pylint is configured for the project, should just be a matter of setting it up in your IDE

python/dataproc_templates/gcs/README.md
ppaglilla: As of now, the templates that support reading csv use sensible defaults for configuration. I agree with your point anyway, let's touch base with the rest of the team on this.

python/dataproc_templates/gcs/gcs_to_spanner.py
ppaglilla: Sure! I'll see about creating an issue to move existing Python templates to this conventions. Resolving this thread.

python/dataproc_templates/gcs/README.md
ppaglilla: Good catch! Let's make the Python template as user-friendly as possible and then backport the improvements to java on a different PR.

python/dataproc_templates/gcs/gcs_to_spanner.py
ppaglilla: Awesome!

python/dataproc_templates/gcs/README.md
shashank-google: @ppaglilla We are for now not testing these templates for text formats "csv" or "json" as it often yields to bugs.
Maybe later down the line we might choose to do it, but not for now.

shashank-google:
@hec10r Bumping this up.
Can you please close the comments and create pull request.

python/dataproc_templates/gcs/gcs_to_spanner.py
hec10r: Good catch. Done in 42fabea

python/dataproc_templates/gcs/gcs_to_spanner.py
hec10r: Done in cc4d05f

python/dataproc_templates/gcs/README.md
hec10r: Done in 947947d

python/dataproc_templates/gcs/gcs_to_spanner.py
hec10r: Is that enough? 9a65512

ppaglilla: Hey! Sorry for the delay reviewing this.

Does the Spanner connector require the Save Mode to be `overwrite` for creating a new table? Or will it create it if we used, for example, `append`?

Just to understand the check we are doing

PR #126, opened 2022-05-25
GCS to JDBC

hhasija:


java/pom.xml
```diff
@@ -242,6 +242,12 @@
       <version>20211205</version>
     </dependency>
 
+    <dependency>
```
ppaglilla, line 245: We shouldn't compile against a JDBC connector for a particular database. It would be better to require the users to add the connector JARs to their job at submit time and parametrize the driver class name. That's what the [existing `JDBCTo*`](https://github.com/GoogleCloudPlatform/dataproc-templates/tree/main/java/src/main/java/com/google/cloud/dataproc/templates/jdbc) connectors do.

java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCSToJDBC.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.execution.datasources.jdbc.JDBCOptions;
+import org.apache.spark.sql.jdbc.JdbcDialects;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class GCSToJDBC implements BaseTemplate {
+  public static final String JDBC_DRIVER = "com.mysql.jdbc.Driver";
```
ppaglilla, line 31: Having the JDBC driver configured this way makes the template only work for MySQL databases. Something like this works for the `GcsToSpanner` template because it only targets Spanner, but here we want the template to work with many different databases.

Like I mentioned in my pom comment, we can refer to the existing `JDBCTo*` templates for this. Those parametrize the driver class and require the user to add the relevant JARs at job submission.

hhasija: Removed that line, and now taking driver name via command line arguments. Thanks for the suggestion.

java/pom.xml
hhasija: Removed the dependency of jar file from pom.xml. Now onwards passing the jar file location via command line argument using export would be required.

ppaglilla (approved):
Great job addressing the comments @hhasija. I haven't been able to test it out myself, but the PR looks good to me code-wise. Approved from my end!

java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCSToJDBC.java
```diff
@@ -0,0 +1,76 @@
...
+
+  @Override
+  public void runTemplate() {
+    try (SparkSession spark = SparkSession.builder().appName("GCS to JDBC").getOrCreate()) {
+      Dataset<Row> dataset =
+          spark.read().format(config.getInputFormat()).load(config.getInputLocation());
+      LOGGER.info("Schema of input Data set is as follows ");
+      dataset.printSchema();
+      LOGGER.info("Actual Data is as follows ");
+      dataset.show();
```
shashank-google, line 54: why dataset.show() ?
This is dump all the data into logs which is undesirable and security risk.

java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCSToJDBC.java
```diff
@@ -0,0 +1,76 @@
...
+
+  public void write(Dataset<Row> dataset) {
+    JdbcDialects.registerDialect(new SpannerJdbcDialect());
+    dataset
+        .write()
+        .format("jdbc")
+        .option(JDBCOptions.JDBC_URL(), config.getJDBCUrl())
+        .option(JDBCOptions.JDBC_TABLE_NAME(), config.getTable())
+        .option(
+            JDBCOptions.JDBC_TXN_ISOLATION_LEVEL(),
```
shashank-google, line 68: I believe, you do not need to specify isolation level property while writing.
please remove and test.

java/src/main/java/com/google/cloud/dataproc/templates/gcs/README.md
```diff
@@ -34,3 +34,23 @@ bin/start.sh \
...
+```
+export JARS=<gcs location for jdbc connector jar>
+
+GCP_PROJECT=<gcp-project-id> \
+REGION=<region>  \
+GCS_STAGING_LOCATION=<gcs-staging-bucket-folder> \
+bin/start.sh \
+-- --template GCSTOJDBC \
+--templateProperty project.id=<gcp-project-id> \
+--templateProperty gcs.jdbc.input.format=<csv | avro | parquet | orc> \
```
shashank-google, line 50: have you tested it for it to works with "csv" ??
Since CSV does not have schema, generally we are not advising it when writing to databases.

java/src/main/resources/template.properties
```diff
@@ -152,6 +152,14 @@ gcs.spanner.output.table=<spanner table id>
 gcs.spanner.output.saveMode=<Append|Overwrite|ErrorIfExists|Ignore>
 gcs.spanner.output.primaryKey=<column[(,column)*] - primary key columns needed when creating the table>
 
+# GCS to JDBC
+gcs.jdbc.input.format=<avro or parquet>
+gcs.jdbc.input.location=<gcs path>
+gcs.jdbc.output.driver=<output jdbc driver>
+gcs.jdbc.output.url=<output jdbc url>
+gcs.jdbc.output.table=<spanner table id>
```
shashank-google, line 160: Why "spanner" table it ??

java/src/main/resources/template.properties
```diff
@@ -152,6 +152,14 @@ gcs.spanner.output.table=<spanner table id>
 gcs.spanner.output.saveMode=<Append|Overwrite|ErrorIfExists|Ignore>
 gcs.spanner.output.primaryKey=<column[(,column)*] - primary key columns needed when creating the table>
 
+# GCS to JDBC
+gcs.jdbc.input.format=<avro or parquet>
```
shashank-google, line 156: orc missing

java/src/main/java/com/google/cloud/dataproc/templates/gcs/README.md
```diff
@@ -34,3 +34,23 @@ bin/start.sh \
...
+--templateProperty gcs.jdbc.input.location=<gcs path> \
+--templateProperty gcs.jdbc.output.driver=<jdbc driver required in single quotes> \
+--templateProperty gcs.jdbc.output.url=<jdbc url along with username and password in single quotes> \
+--templateProperty gcs.jdbc.output.table=<jdbc connection table id> \
+--templateProperty gcs.jdbc.output.saveMode=<Append|Overwrite|ErrorIfExists|Ignore>
+```
```
shashank-google, line 56: Also please add an example launch command with parameter values.

java/src/main/java/com/google/cloud/dataproc/templates/gcs/README.md
```diff
@@ -34,3 +34,23 @@ bin/start.sh \
 --templateProperty gcs.spanner.output.saveMode=<Append|Overwrite|ErrorIfExists|Ignore> \
 --templateProperty gcs.spanner.output.primaryKey=<column[(,column)*] - primary key columns needed when creating the table>
 ```
+
+
+## 3. GCS to JDBC
```
shashank-google, line 39: Readme should specifically call out that JDBC driver needs to be downloaded separately and then uploaded to GCS location.

java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCSToJDBC.java
```diff
@@ -0,0 +1,76 @@
...
+    return new GCSToJDBC(config);
+  }
+
+  @Override
+  public void runTemplate() {
+    try (SparkSession spark = SparkSession.builder().appName("GCS to JDBC").getOrCreate()) {
+      Dataset<Row> dataset =
+          spark.read().format(config.getInputFormat()).load(config.getInputLocation());
+      LOGGER.info("Schema of input Data set is as follows ");
+      dataset.printSchema();
```
shashank-google, line 52: This will print it on standard output and not using logger.
Can you please modify such that it is printed using logger instead.
Maybe something like: `LOGGER.info(dataset.schema());`

shashank-google (changes requested):
Also please add junit tests for the same

java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCSToJDBC.java
hhasija: Removed dataset.show() part. As it was kept initially for only testing purpose.

java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCSToJDBC.java
hhasija: Removed isolation property from the code. And, tested again and the code is working fine.

java/src/main/java/com/google/cloud/dataproc/templates/gcs/README.md
hhasija: Although I tested it with csv and it is working fine but in that case it is taking all the column as string type. So, I guess better is to remove that option. and I did the same.

java/src/main/resources/template.properties
hhasija: changed it to JDBC table.

java/src/main/resources/template.properties
hhasija: Mentioned ORC option as well.

java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCSToJDBC.java
hhasija: Actually these print statements were written for just testing purpose. Hence, removed all those statements to maintain symmetry between all templates.

java/src/main/java/com/google/cloud/dataproc/templates/gcs/README.md
hhasija: Added an example command as well.

java/src/main/java/com/google/cloud/dataproc/templates/gcs/README.md
hhasija: Done.

PR #162, opened 2022-06-21
Text To BigQuery Template

shradha-tyagi:


shradha-tyagi:
CLA issue
Raising a new one

PR #177, opened 2022-06-25
changing python bin start sh file to handle parallel jenkins test

hhasija:


python/bin/start.sh
```diff
@@ -27,7 +27,20 @@ check_required_envvar GCP_PROJECT
...
 check_required_envvar GCS_STAGING_LOCATION
 
-python ${PROJECT_ROOT_DIR}/setup.py bdist_egg --output=$PACKAGE_EGG_FILE
+# Do not rebuild when SKIP_BUILD is specified
+# Usage: export SKIP_BUILD=true
+if [ ! -z "$SKIP_BUILD" ]; then
+
+  #Copy egg file to GCS bucket Staging folder
+  echo_formatted "Copying dist/dataproc_templates_distribution.egg to  staging bucket: ${GCS_STAGING_LOCATION}/dependencies/dataproc_templates_distribution.egg"
+  gsutil cp dist/dataproc_templates_distribution.egg ${GCS_STAGING_LOCATION}/dependencies/dataproc_templates_distribution.egg
```
ppaglilla, line 36: There's no need for us to copy the `.egg` file manually to GCS. The rest of the script references the local copy of the `.egg` file and `gcloud` handles getting the dependencies over to Dataproc.

We could do something like this:

``` bash
if [ -z "$SKIP_BUILD" ]; then
  python ${PROJECT_ROOT_DIR}/setup.py bdist_egg --output=$PACKAGE_EGG_FILE
fi
```

ppaglilla (approved):

PR #190, opened 2022-07-05
Bigquery transformation

hhasija:


java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCStoBigquery.java
```diff
@@ -128,9 +136,17 @@ public void runTemplate() {
           break;
         default:
           throw new IllegalArgumentException(
-              "Currenlty avro, parquet and csv are the only supported formats");
+              "Currently avro, parquet and csv are the only supported formats");
       }
 
+      if (bqTempTable == null || bqTempQuery == null) {
```
franklinWhaite, line 142: So rather than setting default values when bqTempTable or bqTempQuery are not provided, I think it would be cleaner to only apply spark SQL when these are set

franklinWhaite:
Could you please add instruction in the readme explaining how to use the new properties?

java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCStoBigquery.java
hhasija: Done

java/src/main/java/com/google/cloud/dataproc/templates/gcs/README.md
```diff
@@ -18,6 +18,15 @@ bin/start.sh \
 --templateProperty gcs.bigquery.temp.bucket.name=<bigquery temp bucket name>
 ```
 
+There are two optional properties as well with "GCS to BigQuery" Template. Please find below the details :-
+
+```
+--templateProperty gcs.bigquery.temp.table='dataset'
```
franklinWhaite, line 24: is 'dataset' the best name for the table?

hhasija: Done. Have changed it to "table_name" instead of "dataset"

franklinWhaite:
LGTM

franklinWhaite (approved):

java/src/main/java/com/google/cloud/dataproc/templates/gcs/README.md
```diff
@@ -18,6 +18,15 @@ bin/start.sh \
...
 ```
 
+There are two optional properties as well with "GCS to BigQuery" Template. Please find below the details :-
+
+```
+--templateProperty gcs.bigquery.temp.table='table_name' 
+--templateProperty gcs.bigquery.temp.query='select * from global_temp.table_name'
+```
+These properties are responsible for applying some spark sql transformations while loading data into BigQuery.
+The only thing needs to keep in mind is that, the name of dataset and the name of table in the query should match exactly. Otherwise, there would be an error as:- "Table or view not found:"
```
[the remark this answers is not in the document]
hhasija: Corrected.

PR #198, opened 2022-07-18
fixed test failures and dependency

shashank-google:
fixed failing test
updated a dependency

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -127,7 +127,7 @@
     <dependency>
       <groupId>com.google.oauth-client</groupId>
       <artifactId>google-oauth-client</artifactId>
-      <version>1.32.1</version>
+      <version>1.33.3</version>
     </dependency>
 
     <dependency>
--- a/java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCStoBigquery.java
+++ b/java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCStoBigquery.java
@@ -75,7 +75,7 @@ public class GCStoBigquery implements BaseTemplate {
     validateInput();
 
     SparkSession spark = null;
-    LOGGER.info("input format: " + inputFileFormat);
+    LOGGER.info("input format: {}", inputFileFormat);
 
     try {
       spark = SparkSession.builder().appName("GCS to Bigquery load").getOrCreate();
--- a/java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCStoBigquery.java
+++ b/java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCStoBigquery.java
@@ -72,45 +72,9 @@ public class GCStoBigquery implements BaseTemplate {
 
   @Override
   public void runTemplate() {
-    if (StringUtils.isAllBlank(projectID)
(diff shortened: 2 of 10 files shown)


shashank-google:
@hhasija can you please review the same

java/pom.xml
```diff
@@ -127,7 +127,7 @@
     <dependency>
       <groupId>com.google.oauth-client</groupId>
       <artifactId>google-oauth-client</artifactId>
-      <version>1.32.1</version>
+      <version>1.33.3</version>
```
ajayydv, line 130: shall we make this as maven variable?

java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCStoBigquery.java
```diff
@@ -72,45 +72,9 @@ public GCStoBigquery() {
...
-        GCS_OUTPUT_DATASET_NAME,
-        bigQueryDataset,
-        GCS_OUTPUT_TABLE_NAME,
-        bigQueryTable,
-        GCS_BQ_INPUT_FORMAT,
-        inputFileFormat,
-        GCS_BQ_LD_TEMP_BUCKET_NAME,
-        bqTempBucket);
-
     LOGGER.info("input format: " + inputFileFormat);
```
ajayydv, line 78: Lets reformat this for lazy initialization in logger. i.e
LOGGER.info("input format: {}", inputFileFormat);

java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCStoBigquery.java
```diff
@@ -160,4 +124,44 @@ public void runTemplate() {
       }
     }
   }
+
```
ajayydv, line 127: Please add a javadoc comment for function.

java/src/main/java/com/google/cloud/dataproc/templates/hive/HiveToGCS.java
```diff
@@ -122,4 +94,35 @@ public void runTemplate() {
     LOGGER.info("HiveToGcs job completed.");
     spark.stop();
   }
+
+  void validateInput() {
```
ajayydv, line 98: Shall we consider moving validateInput to BaseTemplate? I understand that input parameters are different for each parameter but its worth consolidating them in base class via var args or lambada function. We can create a seperate jira for this if required.

ajayydv (changes requested):
Thanks for working on this Shashank. Added few comments. LGTM with those addressed.

java/src/main/java/com/google/cloud/dataproc/templates/hive/HiveToGCS.java
shashank-google: yeah we should create a separate git issue for code refactoring.

java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCStoBigquery.java
shashank-google: Fixing it. FYI, it is not a change made in this pull request.

Issue #206, opened 2022-07-25
[java] Tech Debt - Add JUnit tests for templates


shashank-google:
Add JUnit tests for below templates, you may use [HiveToGCSTest.java](https://github.com/GoogleCloudPlatform/dataproc-templates/blob/main/java/src/test/java/com/google/cloud/dataproc/templates/hive/HiveToGCSTest.java) as example.
Goal is write unit test for all the business logic before spark session is created.

[JDBCToGCS.java](https://github.com/GoogleCloudPlatform/dataproc-templates/blob/main/java/src/main/java/com/google/cloud/dataproc/templates/jdbc/JDBCToGCS.java)
[JDBCToBigQuery.java](https://github.com/GoogleCloudPlatform/dataproc-templates/blob/main/java/src/main/java/com/google/cloud/dataproc/templates/jdbc/JDBCToBigQuery.java)
[SpannerToGCS.java](https://github.com/GoogleCloudPlatform/dataproc-templates/blob/main/java/src/main/java/com/google/cloud/dataproc/templates/databases/SpannerToGCS.java)
[DataplexGCStoBQ.java](https://github.com/GoogleCloudPlatform/dataproc-templates/blob/main/java/src/main/java/com/google/cloud/dataproc/templates/dataplex/DataplexGCStoBQ.java)

poojabasker20:
JUnit tests for [JDBCToGCS.java](https://github.com/GoogleCloudPlatform/dataproc-templates/blob/main/java/src/main/java/com/google/cloud/dataproc/templates/jdbc/JDBCToGCS.java) and [JDBCToBigQuery.java](https://github.com/GoogleCloudPlatform/dataproc-templates/blob/main/java/src/main/java/com/google/cloud/dataproc/templates/jdbc/JDBCToBigQuery.java) have been added as part of issue https://github.com/GoogleCloudPlatform/dataproc-templates/issues/464

hhasija:
Closing out this issue, as different folks are working on different unit tests. Thus, created different issues separately for different templates :-

https://github.com/GoogleCloudPlatform/dataproc-templates/issues/464

https://github.com/GoogleCloudPlatform/dataproc-templates/issues/466

https://github.com/GoogleCloudPlatform/dataproc-templates/issues/530

PR #234, opened 2022-08-04
Template/hbasetogcs - Unit Tests HbaseToGCS Java

anish97IND:


java/src/main/java/com/google/cloud/dataproc/templates/hbase/HbaseToGCS.java
```diff
@@ -61,20 +61,26 @@ public void runTemplate() {
...
+      optionsMap.put(HBaseTableCatalog.tableCatalog(), catalogue.replaceAll("\\s", ""));
+      // Read from HBase
+      dataset =
+          spark
+              .read()
+              .format("org.apache.hadoop.hbase.spark")
+              .options(optionsMap)
+              .option("hbase.spark.use.hbasecontext", "false")
+              .load();
+    } catch (NoClassDefFoundError error) {
```
shashank-google, line 76: Adding try catch just for "test case" is not an option.
We don't need to junit test spark session and onwards code.
Look at HiveToGCSTest.java to create test accordingly.

shashank-google (changes requested):
Adding try catch is not acceptable for junit tests.
Follow the lead of below github issue to fix test in similar way.
https://github.com/GoogleCloudPlatform/dataproc-templates/issues/206

shashank-google (approved):

PR #252, opened 2022-08-17
fixing bug in filter

franklinWhaite:
Fixed the following issues:

1. In the Dataplex template, when comparing partitions in Dataplex vs Partitions in BigQuery the name of columns used to compare between partitions was hardcoded as `t2.id`, therefore the template would fail if the data lacked a partition key named `id`. 
2. A string literal was being used for Spark format for BigQuery in several locations across the project, changed this into a constant.


ajayydv (approved):
LGTM, thanks for the patch Franklin!!

PR #257, opened 2022-08-21
adding spark sql based transformations for python templates

hhasija:


NiloFreitas:
Could you add unit tests to covert the added properties?

shashank-google:
@hhasija can you please update it with the latest so that it can be merged.

PR #259, opened 2022-08-23
JDBC to GCS fix

### LINKED ISSUE #258: [python][bug] JDBC to GCS csv headers not populating
CSV output using jdbc to gcs template is not populating headers.
Fix Line 195 in code change .option(constants.FORMAT_CSV, True) to .option(constants.HEADER, True)

surjits254:
added header=true for csv writre

closes #258


NiloFreitas (approved):
LGTM

PR #273, opened 2022-09-02
Automation Enhancement and Bug fix-: Hbase To GCS  Python Template

anish97IND:


python/bin/start.sh
```diff
@@ -62,6 +62,30 @@ fi
 if [ -n "${PY_FILES}" ]; then
   OPT_FILES="${OPT_PY_FILES},${PY_FILES}"
 fi
+#if Hbase catalog is passed, then required hbase dependency are copied to staging location and added to jars
+if [ -n "${CATALOG}" ]; then
```
surjits254, line 66: add a variable for skipping the docker build if user wants to reuse already available image

python/bin/start.sh
```diff
@@ -62,6 +62,30 @@ fi
...
+  echo "Downloading Hbase jar dependency"
+  wget https://repo1.maven.org/maven2/org/apache/hbase/hbase-client/2.4.12/hbase-client-2.4.12.jar
+  wget https://repo1.maven.org/maven2/org/apache/hbase/hbase-shaded-mapreduce/2.4.12/hbase-shaded-mapreduce-2.4.12.jar
+  wget https://repo1.maven.org/maven2/org/apache/htrace/htrace-core4/4.2.0-incubating/htrace-core4-4.2.0-incubating.jar
+  gsutil copy hbase-client-2.4.12.jar ${GCS_STAGING_LOCATION}/hbase-client-2.4.12.jar
+  gsutil copy hbase-shaded-mapreduce-2.4.12.jar ${GCS_STAGING_LOCATION}/hbase-shaded-mapreduce-2.4.12.jar
+  gsutil copy htrace-core4-4.2.0-incubating.jar ${GCS_STAGING_LOCATION}/htrace-core4-4.2.0-incubating.jar
+  echo "Passing downloaded dependency jars"
+  OPT_JARS="${OPT_JARS},${GCS_STAGING_LOCATION}/hbase-client-2.4.12.jar,${GCS_STAGING_LOCATION}/hbase-shaded-mapreduce-2.4.12.jar,${GCS_STAGING_LOCATION}/htrace-core4-4.2.0-incubating.jar,file:///usr/lib/spark/external/hbase-spark.jar"
```
surjits254, line 75: do we need file:///usr/lib/spark/external/hbase-spark-protocol-shaded.jar lib?

python/bin/start.sh
```diff
@@ -62,6 +62,30 @@ fi
...
+  rm hbase-client-2.4.12.jar
+  rm hbase-shaded-mapreduce-2.4.12.jar
+  rm htrace-core4-4.2.0-incubating.jar
+fi
+
+if [ -n "${HBASE_SITE_PATH}" ]; then
+  #Copy the hbase-site.xml to docker context
+  cp $HBASE_SITE_PATH .
+  export HBASE_SITE_NAME=`basename $HBASE_SITE_PATH`
+  docker build -t "${IMAGE}" -f dataproc_templates/hbase/Dockerfile --build-arg HBASE_SITE_NAME=${HBASE_SITE_NAME} .
```
surjits254, line 85: I think there is issue with docker build from Mac M1 Pro, can you research around this and add --platform which works consitently?

I added docker build --platform=linux/amd64 to make it work otherwise I get exec /entrypoint.sh: exec format error

python/dataproc_templates/hbase/README.md
```diff
@@ -88,12 +91,15 @@ optional arguments:
...
 ```
+
 export GCP_PROJECT=<project_id>
 export SUBNET=<subnet>
 export REGION=<region>
-export GCS_STAGING_LOCATION=<gcs-staging-bucket-folder> 
+
+#For manual process JARS environment variable need to be set. Not required for automated process
 export JARS="gs://<your_bucket_to_store_dependencies>/hbase-client-2.4.12.jar, \
```
surjits254, line 103: can we add htrace-core4-4.2.0-incubating.jar dep in the manual process section as well?

python/dataproc_templates/hbase/README.md
```diff
@@ -5,7 +5,9 @@ Template for reading files from Hbase and writing to Google Cloud Storage. It su
 ## Requirements
```
surjits254, line 5: can we add docker as a requirement. Also for pushing images from local to container register I had to run gcloud auth configure-docker can we capture this as well?

python/dataproc_templates/hbase/README.md
```diff
@@ -41,7 +43,8 @@ Template for reading files from Hbase and writing to Google Cloud Storage. It su
 
 ## Required JAR files
```
surjits254, line 44: update this section with htrace dep as well

surjits254 (changes requested):

python/bin/start.sh
anish97IND: No its not required. hbase-spark is sufficient.

python/dataproc_templates/hbase/README.md
anish97IND: Thanks for the suggestion. Since, this jar is specific to few of the Hbase version, early- 2.x , hence, this is captured in the comment section. I can add an optional jar export example for htrace-core4-4.2.0-incubating.jar.

surjits254: yes that would be great

python/bin/start.sh
anish97IND: I think it is mentioned in the read me section by the template creator, however its not required. I will remove it from there to make it clearer.

python/bin/start.sh
anish97IND: Thanks for bringing this. Yeah correct this is specific to Mac. Hence, it might be a separate enhancement. For now this PR is more focused on automating the process of image build. However, I will capture about this "--platform"  in Read me.

anish97IND: Also I did look at it, looks like this is automatically set by the docker itself. This issue might be due to wrong installation or installation problem specific to Mac. Added link in the new commit. Ref https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

python/dataproc_templates/hbase/README.md
anish97IND: The required setup for connection in between local and cloud is a predefined step. Addressed it in the read me.

python/bin/start.sh
anish97IND: Added the functionality for the same, however the step highlighted here is only for jar dependency and not for the image build process. Please refer HBASE_SITE_PATH if statement.

python/bin/start.sh
surjits254: can you remove it from the manual process?

surjits254 (changes requested):

python/bin/start.sh
anish97IND: Oh yes, I missed that out.

surjits254 (approved):
please resolve sync your branch with main and merge

Issue #280, opened 2022-09-07
[java] Update dependency 


shashank-google:
Security venerability, update following dependency to following.
Ensure build and tests are working.

<dependency>
  <groupId>org.apache.hadoop</groupId>
  <artifactId>hadoop-common</artifactId>
  <version>3.2.4</version>
</dependency>

ishu-thakur:
@shashank-google i want to work on this if this issue is still open ?

shashank-google:
@ishu-thakur it was fixed already

PR #282, opened 2022-09-08
update hadoop version to fix vul

kvaithin:
Fixes: #280

Version update for hadoop-common to fix vulnerability.


kvaithin:
Could ignore, seem this issue is already addressed.

PR #292, opened 2022-09-14
[New Template][Java] Cassandra to GCS 

anish97IND:


java/pom.xml
```diff
@@ -115,7 +115,17 @@
       <artifactId>spark-sql-kafka-0-10_${scala.binary.version}</artifactId>
       <version>${spark.version}</version>
     </dependency>
-
+    <!-- https://mvnrepository.com/artifact/com.datastax.spark/spark-cassandra-connector -->
+    <dependency>
+      <groupId>com.datastax.spark</groupId>
+      <artifactId>spark-cassandra-connector_2.12</artifactId>
```
shashank-google, line 121: use variable scala.binary.version

java/pom.xml
```diff
@@ -115,7 +115,17 @@
...
+      <version>3.1.0</version>
+    </dependency>
+    <dependency>
+      <groupId>com.github.jnr</groupId>
+      <artifactId>jnr-posix</artifactId>
```
shashank-google, line 126: are we sure that this dependency is really needed ?
What is the error without it.

java/src/main/java/com/google/cloud/dataproc/templates/databases/CassandraToGCS.java
```diff
@@ -0,0 +1,103 @@
...
+    validateInput();
+    SparkSession spark =
+        SparkSession.builder()
+            .appName("Spark CassandraToGCS Job")
+            .withExtensions(new CassandraSparkExtensions())
+            .config(
+                "spark.sql.catalog.dc", "com.datastax.spark.connector.datasource.CassandraCatalog")
+            .config("spark.sql.catalog.dc.spark.cassandra.connection.host", cassandraHost)
+            .getOrCreate();
+    Dataset dataset = spark.sql("SELECT * FROM dc" + "." + keyspace + "." + table);
```
shashank-google, line 58: Use String.format instead of string concatenation.
is it appropriate to hard code `dc` schema here ?
Is it possible that there is no `keyspace` and only a tablename ?

Why not allow user to input the cassandra sql query instead, like jdbcToGCS templates ?

java/src/main/resources/template.properties
```diff
@@ -145,6 +145,14 @@ spanner.gcs.output.gcs.format=<avro|csv|parquet|json|orc>
 spanner.gcs.temp.table=
 spanner.gcs.temp.query=
 
+# Cassandra To GCS properties
```
shashank-google, line 148: Values MUST not be specified against keys.

java/src/main/java/com/google/cloud/dataproc/templates/databases/CassandraToGCS.java
```diff
@@ -0,0 +1,103 @@
...
+    dataset.write().format(outputFileFormat).mode(gcsSaveMode).save(gcsWritePath);
+  }
+
+  public void validateInput() {
```
shashank-google, line 62: Checks need to be more comprehensive. Not null is not sufficient.
Consider using Jakarta validataion framework as used in `JDBCToGCSConfig.java`

shashank-google (changes requested):

shashank-google:
@hhasija can you please review as well and add your review comment if any.

hhasija:
> @hhasija can you please review as well and add your review comment if any.

Sure @shashank-google

java/pom.xml
anish97IND: Yeah, initially I did not have it, that when I got a ClassNotFound exception within this package. The exception was within an internal class of Cassandra connector. Adding this package, solved the issue.

java/src/main/java/com/google/cloud/dataproc/templates/databases/CassandraToGCS.java
anish97IND: "dc" is a catalog here (or connection name to cassandra) (and not schema) and is related to establishing connection with a cassandra node. That node will be first point of contact to establish connection with the cluster and then connector will automatically connect to other nodes within the cassandra ring. Since, we are trying to connect to a single cluster at a point of time, hence, having "dc" hardcoded might not be a problem. `keyspace` in Cassandra basically means schema in database. Each `table` exist within one of the `keyspace`.  Since Cassandra Extention is loaded in this pipeline for SparkSQL, CQL type commands can be passed along with additional optimisation it brings. I was looking at first doing a plan lift and shift, however, I can add an additional feature for SQL based migrations.

shashank-google (approved):

java/README.md
```diff
@@ -5,6 +5,7 @@
 Please refer to the [Dataproc Templates (Java - Spark) README](java/README.md)  for more information
 * [HiveToBigQuery](/java/src/main/java/com/google/cloud/dataproc/templates/hive/README.md)
 * [HiveToGCS](/java/src/main/java/com/google/cloud/dataproc/templates/hive/README.md)
+* [CassandraToGCS](java/src/main/java/com/google/cloud/dataproc/templates/databases/README.md)
```
shashank-google, line 8: need to add the same on /README.MD as well.

shashank-google (approved):
@anish97IND 
1. Please add the link in top readme to this template.
2. resolve merge conflicts and merge the same.

PR #372, opened 2022-10-22
addition of hardening statistics in read me file

hhasija:


NiloFreitas (approved):

shashank-google:
IMHO the information here is far too cryptic and impossible to understand by an outside reader.
There will be questions like:

- What is meant by hardening ?
- What was table schema ?
- How many files were in GCS, what was avg file size ?
- How many spark executors were used, machine type ?
- How many BQ slots were used ?
- What BQ features were used (like Transactions, Generated Columns etc etc) ?

In other words, it cannot be just the stats.
We need to publish in full detail, how can someone re-run same tests and re-produce same results.

PR #395, opened 2022-11-11
Updated Google-auth python package

shubhamgoogle:
Due to recent changes in the google-auth-library-python-oauth module python was not able to connect to GCS and thus was not able to save audit data as csv in GCS.

Ref: https://github.com/googleapis/google-auth-library-python-oauthlib/pull/240/files#diff-63cdf7d9059947cf2f03b72a8137b8a1901de52f685705c251886d58a75b9381L8

Updates Google-auth library to required package to fix the issue.


shashank-google:
@shubhamgoogle 
Just to understand it better, why are our integration tests working without this change.

shubhamgoogle:
Hi @shashank-google, This part of saving audit table is not covered in base templates.

shashank-google (approved):

PR #400, opened 2022-11-21
changing links to deep links in read me files

hhasija:


shashank-google (approved):

shashank-google:
"Github" is not rendering to scroll automatically to right heading when opened in same tab. 
But when opened in new tab it auto scrolls. So we can go ahead with the change.

@hhasija please resolve the merge conflicts and then merge this

PR #429, opened 2022-12-05
Issue 405: [Python/Java Templates] Add JDBC Data Source Attribute: fetchsize

nj1973:
As far as we are aware the JDBC data source fetchsize attribute is only important to set when Oracle is the source. This is because the Oracle JDBC driver has a low default (10). I've added the attribute to JDBC templates but only referenced setting it in Oracle sections of READMEs and in Oracle based notebooks.

Changes:

- Add fetchsize as optional attribute to Python templates:
-- jdbc_to_bigquery.py
-- jdbc_to_gcs.py
-- jdbc_to_jdbc.py
- Add fetchsize to Python unit tests
- Add fetchsize as optional attribute to Java templates
-- JDBCToBigQuery.java
-- JDBCToGCS.java
-- JDBCToSpanner.java
- Add fetchsize=200 to Oracle based Python notebooks:
-- OracleToBigQuery_notebook.ipynb + README.md
-- OracleToSpanner_notebook.ipynb

I've manually tested and run the branch through Jenkins projects.


java/src/main/java/com/google/cloud/dataproc/templates/util/TemplateConstants.java
```diff
@@ -124,6 +126,7 @@ public interface TemplateConstants {
   String JDBC_TO_SPANNER_JDBC_URL = "jdbctospanner.jdbc.url";
 
   String JDBC_TO_SPANNER_JDBC_DRIVER_CLASS_NAME = "jdbctospanner.jdbc.driver.class.name";
+  String JDBC_TO_SPANNER_JDBC_FETCH_SIZE = "jdbctospanner.jdbc.fetchsize";
```
shashank-google, line 129: corresponding change in template.properties file is missing
Also please use a reasonable default fetch size if it makes sense (or zero) in template.properties

java/src/main/java/com/google/cloud/dataproc/templates/jdbc/JDBCToGCS.java
```diff
@@ -63,6 +63,10 @@ public void runTemplate() {
       jdbcProperties.put(JDBCOptions.JDBC_NUM_PARTITIONS(), config.getJdbcSQLNumPartitions());
     }
 
+    if (StringUtils.isNotBlank(config.getJdbcFetchSize())) {
```
shashank-google, line 66: if you are putting a default value against it then "if condition" is not required and can be removed.

shashank-google (changes requested):

notebooks/oracle2spanner/README.md
```diff
@@ -43,7 +43,11 @@ Below configurations are required before proceeding further.
 * `SPANNER_OUTPUT_MODE`: <Append | Overwrite>
 * `SPANNER_INSTANCE` : Cloud Spanner instance name
 * `SPANNER_DATABASE` : Cloud Spanner database name
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
nj1973: Fixed, thanks

java/src/main/java/com/google/cloud/dataproc/templates/util/TemplateConstants.java
nj1973: I've added entries to this file. I've used default of zero. I chose zero so that each driver will choose it's own value. As far as I know only Oracle has an unsuitable default. I didn't want to force my own default on other JDBC drivers.

For my education: what is this file used for? I didn't have any failing tests in Jenkins without this edit.

java/src/main/java/com/google/cloud/dataproc/templates/jdbc/JDBCToGCS.java
nj1973: I've decided not to put a value against the property in template.properties. The JDBC reader has a default of zero so I thought it was fine for me to not set it at all if the user hasn't provided a value.

java/src/main/java/com/google/cloud/dataproc/templates/util/TemplateConstants.java
nj1973: I found where the file is used. Is my understanding correct that it is used to provide default values?

shashank-google (approved):

java/src/main/java/com/google/cloud/dataproc/templates/util/TemplateConstants.java
shashank-google: Historically we were using a different cli parser, which forced us to have cli parameter name declared in template.properties
Down the line it was changed to current parser, which has removed that restriction.
Currently we are still following that convention. 
Which I think is ok, but if we want to change I am open to discuss it.

PR #440, opened 2022-12-09
436 documentationbug java jdbctobq example submission

arindam-google:
Fixed the issue for Java JDBC to BQ by changing the `Readme.md` by changing the template property `jdbctobq.temp.gcs.bucket` as `gcs temp bucket name` instead of `gcs bucket path`


hhasija (approved):
LGTM

Issue #157, opened 2022-06-17
[NewTemplate][Java] GCSToBigTable


shashank-google:
similar to python template

shashank-google:
HI @anish97IND If you are not working on this gitissue, can you please un-assign yourself.

anish97IND:
Hi @shashank-google , I have completed the coding part but was facing issue with a dependency. Hence, was unable to raise PR for it. However, I will un-assign myself, will revert back if I find a triage.

PR #442, opened 2022-12-12
GCS to BigTable new template in Java

balajiss2:
#157 

This is a new template for GCS to BigTable in Java. This has capability to load CSV, AVRO and Parquet Files into Bigtable

https://github.com/GoogleCloudPlatform/dataproc-templates/issues/157


java/src/main/resources/template.properties
```diff
@@ -161,6 +161,18 @@ gcs.bigquery.output.dataset=
 gcs.bigquery.output.table=
 gcs.bigquery.temp.bucket.name=
 
+#GCStoBigTable properties
+# supported formats avro,parquet,csv
+gcs.bigtable.input.format=<csv|parquet|avro>
+#gcs file location
+gcs.bigtable.input.location=<input file location>
```
vanshaj-bhatia, line 168: Remove whatever is after =
These values are used for validation. Line 166 is the only correct usage

java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCStoBigTable.java
```diff
@@ -0,0 +1,173 @@
...
+
+    SparkSession spark = null;
+    LOGGER.info("input format: {}", inputFileFormat);
+
+    try {
+      spark = SparkSession.builder().appName("GCS to Bigtable load").getOrCreate();
+
+      Dataset<Row> inputData = null;
+
+      switch (inputFileFormat) {
```
vanshaj-bhatia, line 68: Why do we need this switch-case?

balajiss2: The template has the capability to process CSV, AVRO and Parquet files. The switch case is there in order to send the right parameters for the dataframe read operator. For instance, for CSV, we need to provide details on header and all but for parquet,avro there is no need to pass those values. That is the need for the switch case

java/src/main/resources/template.properties
balajiss2: made the changes and committed

java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCStoBigTable.java
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  @Override
+  public void runTemplate() {
+    validateInput();
+
+    SparkSession spark = null;
+    LOGGER.info("input format: {}", inputFileFormat);
+
+    try {
```
vanshaj-bhatia, line 63: remove try-catch

balajiss2: removed try-catch

vanshaj-bhatia:
LGTM!
Please test the template before merging and proceed with the next steps.

Issue #346, opened 2022-10-03
[notebook] [bug] Hardcoded WORKING_DIRECTORY


shashank-google:
All notebooks have hardcoded working directory.
It needs to be dynamically computed as user can potentially checkout in a different sub-directory.

anshumanwins:
I have created  the PR for this issue . 
It solves the basic problem of hardcoded working directory however for cases where notebooks are run on spark cluster rather than locally , we still need to specify the path manually.
 
Kindly refer to PR comments for more details

PR #465, opened 2022-12-23
#Issue 346: Workaround for hardcoded working dir 

anshumanwins:
[Issue Link](  https://github.com/GoogleCloudPlatform/dataproc-templates/issues/346
)

For this I have used os.getcwd() in places where I found the parameter 'WORKING_DIRECTORY'  hardcoded in each notebooks.  
Earlier path used in notebooks were  like this
` WORKING_DIRECTORY = "/home/jupyter/dataproc-templates/python/"`
What I did is used os.getcwd() to get current directory and naviagate to python folder using that.  This will work in case we are running it locally however when using PySpark hive cluster os.getcwd() will return '/' hence we need to manually set the directory path . Thus i have written this as a comment in notebook.

`cur_path = Path(os.getcwd())`
`WORKING_DIRECTORY = os.path.join(cur_path.parent.parent ,'python')`

`# If the above code doesn't fetches the correct path please`
`# provide complete path to python folder in your dataproc `
`# template repo which you cloned `

`# WORKING_DIRECTORY = "/home/jupyter/dataproc-templates/python/"`
`print(WORKING_DIRECTORY)`

This change was done in all 7 notebooks.

### DIFF
--- a/notebooks/generic_notebook/vertex_pipeline_pyspark.ipynb
+++ b/notebooks/generic_notebook/vertex_pipeline_pyspark.ipynb
@@ -130,7 +130,8 @@
     "import google.cloud.aiplatform as aiplatform\n",
     "from kfp import dsl\n",
     "from kfp.v2 import compiler\n",
-    "from datetime import datetime"
+    "from datetime import datetime\n",
+    "from pathlib import Path"
    ]
   },
   {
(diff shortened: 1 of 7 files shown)


surjits254 (commented):
LGTM

NiloFreitas (approved):
LGTM

PR #482, opened 2023-01-09
feat: Add csv as a valid input format for GCSTOJDBC Java template

nj1973:
Added CSV to valid input formats and added relevant code to read section of template.

### DIFF
--- a/java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCSToJDBC.java
+++ b/java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCSToJDBC.java
@@ -15,6 +15,13 @@
  */
 package com.google.cloud.dataproc.templates.gcs;
 
+import static com.google.cloud.dataproc.templates.util.TemplateConstants.GCS_JDBC_AVRO_FORMAT;
+import static com.google.cloud.dataproc.templates.util.TemplateConstants.GCS_JDBC_CSV_FORMAT;
+import static com.google.cloud.dataproc.templates.util.TemplateConstants.GCS_JDBC_CSV_HEADER;
+import static com.google.cloud.dataproc.templates.util.TemplateConstants.GCS_JDBC_CSV_INFER_SCHEMA;
+import static com.google.cloud.dataproc.templates.util.TemplateConstants.GCS_JDBC_ORC_FORMAT;
+import static com.google.cloud.dataproc.templates.util.TemplateConstants.GCS_JDBC_PRQT_FORMAT;
+
 import com.google.cloud.dataproc.dialects.SpannerJdbcDialect;
 import com.google.cloud.dataproc.templates.BaseTemplate;
 import com.google.cloud.dataproc.templates.util.PropertyUtil;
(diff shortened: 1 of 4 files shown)


java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCSToJDBC.java
```diff
@@ -47,13 +54,39 @@ public static GCSToJDBC of(String... args) {
...
+                .option(GCS_JDBC_CSV_HEADER, true)
+                .option(GCS_JDBC_CSV_INFER_SCHEMA, true)
+                .load(config.getInputLocation());
+        break;
+      case GCS_JDBC_AVRO_FORMAT:
+      case GCS_JDBC_ORC_FORMAT:
+        inputData = spark.read().format(config.getInputFormat()).load(config.getInputLocation());
+        break;
+      case GCS_JDBC_PRQT_FORMAT:
+        inputData = spark.read().parquet(config.getInputLocation());
```
shashank-google, line 80: I would prefer if we can use same code as for Avro and ORC. 
Why are we using a different code for parquet ?
`inputData = spark.read().format(config.getInputFormat()).load(config.getInputLocation());`

java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCSToJDBC.java
```diff
@@ -47,13 +54,39 @@ public static GCSToJDBC of(String... args) {
...
+    Dataset<Row> inputData = null;
+
+    switch (config.getInputFormat()) {
+      case GCS_JDBC_CSV_FORMAT:
+        inputData =
+            spark
+                .read()
+                .format(config.getInputFormat())
+                .option(GCS_JDBC_CSV_HEADER, true)
+                .option(GCS_JDBC_CSV_INFER_SCHEMA, true)
```
shashank-google, line 72: If we are now entertaining CSV as well, then we should also add "delimiter" as input too.
In addition there can be many more other configurable options like (dateFormat, nullValue, escape etc).
This is open for discussion.

Taking clues from respective Dataflow template.
https://cloud.google.com/dataflow/docs/guides/templates/provided-batch#gcs_text_to_cloud_spanner

shashank-google (changes requested):

java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCSToJDBC.java
nj1973: I followed what had been done previously in the `GCSToBigQuery.java` template. That one has hardcoded options for `inferSchema` and `header`. So does the equivalent Python `gcs_to_jsbc.py` template.

java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCSToJDBC.java
nj1973: In keeping with trying to be consistent with other templates I only added the same two hardcoded options. I agree these these should be configurable and we should look to add more control. But I think that is something to be implemented across all CSV related templates and not just this one.
I believe it should be an issue in its own right.

nj1973: Agreed verbally that there should be another issue raised for full CSV option control across all relevant templates.

shashank-google: +1

shashank-google (approved):

PR #488, opened 2023-01-12
Removed placeholder values in template.properties

poojabasker20:
Will remove spark.sql.warehouse.dir property from template.properties as part of issue https://github.com/GoogleCloudPlatform/dataproc-templates/issues/487


shashank-google (approved):
LGTM

PR #492, opened 2023-01-14
docs: README Files for PostgreSQL Targets Should Mention reWriteBatchedInserts

nj1973:
Added reWriteBatchedInserts=True to any examples of commands using PostgreSQL as a JDBC target.
Added recommendation to include reWriteBatchedInserts=True in PostgreSQL URLS to README files.


surjits254 (approved):
LGTM

PR #435, opened 2022-12-08
Added Snowflake To GCS - Java Blogpost Link

vanshaj-bhatia:
This PR closes #314

PR #493, opened 2023-01-16
feat: 480 python gcstojdbc template needs numpartitions to match java template

tanyarw:


python/dataproc_templates/gcs/gcs_to_jdbc.py
```diff
@@ -94,6 +94,13 @@ def parse_args(args: Optional[Sequence[str]] = None) -> Dict[str, Any]:
             default=1000,
             help='JDBC output batch size'
         )
+        parser.add_argument(
+            f'--{constants.GCS_JDBC_NUMPARTITIONS}',
+            dest=constants.GCS_JDBC_NUMPARTITIONS,
+            required=False,
+            default="10",
+            help='The maximum number of partitions that can be used for parallelism in table writing. Default set to 10'
```
nj1973, line 102: Why default to 10? The idea of this issue was to make the Python template consistent with the Java template.
As far as I can tell the Java template has no default value and has no effect if it is not specified.
I think, if the user does not specify this attribute then it should remain set to `None` and have no effect.

tanyarw: I picked up the default value inspired from python's JDBCTOJDBC, I should've compared it with the Java template too.

nj1973: I see, that makes sense. What you've done may be correct then but we should discuss because it's hard to be consistent when other templates are inconsistent :-)

tanyarw: Keeping the attribute `None` throws null pointer exception. Instead, I've assigned it the same value as read partitions

nj1973: I've tested this manually and it looks good to me. I'll resolve this conversation.

nj1973:
Do integrations tests need running in Jenkins for this branch?

tanyarw:
> Do integrations tests need running in Jenkins for this branch?

That's been checked too (Build #435)

nj1973 (approved):
Looks good to me

nj1973:
Looks like there's a merge conflict to resolve. I think you'll also need to change the title of the PR to start with either "feat:" or "refactor:" for the new commit checker.

tanyarw:
> Looks like there's a merge conflict to resolve. I think you'll also need to change the title of the PR to start with either "feat:" or "refactor:" for the new commit checker.

Thanks for the heads-up. Resolved it

nj1973 (approved):
LGTM

PR #504, opened 2023-01-19
Ddl extractor

shubhamgoogle:
Raising PR for Hive DDL extractor related to issue #398 
fix: #398


shubhamgoogle:
Issue: https://github.com/GoogleCloudPlatform/dataproc-templates/issues/218

shashank-google:
@shubhamgoogle this is not right commit message.
DDL Extractor is not a bug fix but it is new feature imho
@ankuljain09 what do you suggest to do in a case like this ?

ankuljain09:
Yes @shashank-google Agreed!

shashank-google:
> Yes @shashank-google Agreed!

@ankuljain09 what is right way to update this.
So that this pull request would not go in as fix in release notes but as feature instead ?

ankuljain09:
> > Yes @shashank-google Agreed!
> 
> @ankuljain09 what is right way to update this. So that this pull request would not go in as fix in release notes but as feature instead ?

Another commit on the same PR with a **feat:** should acknowledge this commit as a feature. It may also include the previous commit as a patch in the ChangeLog, In that case we'll have to edit the release notes before the publish.

shashank-google:
> Another commit on the same PR with a **feat:** should acknowledge this commit as a feature. It may also include the previous commit as a patch in the ChangeLog, In that case we'll have to edit the release notes before the publish.

LGTM

shubhamgoogle:
Added new feat: comment.

python/dataproc_templates/hive/util/hive_ddl_extractor.py
```diff
@@ -0,0 +1,86 @@
...
+            )
+        
+               
+        def get_ddl(hive_database, table_name):
+            return spark.sql(f"SHOW CREATE TABLE {hive_database}.{table_name}").rdd.map(lambda x: x[0] + ";").collect()[0]
+
+        ct = datetime.now().strftime("%m-%d-%Y %H.%M.%S")
+        output_path = gcs_output_path+"/"+hive_database+"/"+str(ct)
+
+        tables_names = spark.sql(f"SHOW TABLES IN {hive_database}").select("tableName")
```
NiloFreitas, line 81: Hey @shubhamgoogle , I changed the code to simplify the arguments, fixed some minor issues, and improved documentation.

Only thing open here is the query, which I removed the "AS SERDE" suffix, because I got this error when testing the template:

`pyspark.sql.utils.AnalysisException: `default`.`empdatanew` is a Spark data source table. Use `SHOW CREATE TABLE` without `AS SERDE` instead`

Could you double check if we really need this?

shubhamgoogle: Hi @NiloFreitas , If the table is created by HIVE and we do not use "as serde" using spark this is the output we get:

|CREATE TABLE `default`.`employees` (
  `eid` INT,
  `name` STRING,
  `position` STRING,
  `dept` STRING)
USING text
LOCATION 'hdfs://hive-cluster-m/user/hive/warehouse/partitiondb.db/employee'
TBLPROPERTIES (
  'bucketing_version' = '2',
  'transient_lastDdlTime' = '1658416638')
  
 Now if we use "as serde" in show create command using spark we get:
 
 |CREATE EXTERNAL TABLE `default`.`employees`(
  `eid` INT,
  `name` STRING,
  `position` STRING,
  `dept` STRING)
ROW FORMAT SERDE 'org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe'
**WITH SERDEPROPERTIES (
  'serialization.format' = ',',
  'field.delim' = ',')**
STORED AS
  INPUTFORMAT 'org.apache.hadoop.mapred.TextInputFormat'
  OUTPUTFORMAT 'org.apache.hadoop.hive.ql.io.HiveIgnoreKeyTextOutputFormat'
LOCATION 'hdfs://hive-cluster-m/user/hive/warehouse/partitiondb.db/employee'
TBLPROPERTIES (
  'bucketing_version' = '2',
  'transient_lastDdlTime' = '1658416638')
  
As you can see we will not get serde properties like delimiter(which is important for text files) if we do not use "as serde".

"As serde" will not work if the table is created using spark instead, but as it is HIVE DDL extractor I thought we should add it. We can definitely remove it and mention in the document this limitation.

NiloFreitas: Ok. I have put it back and documented. LGTM feel free to merge

PR #509, opened 2023-01-19
fix: CassandraToBQ catalog name does not default to casscon

poojabasker20:


vanshaj-bhatia:
@poojabasker20 Can we also remove these from TemplateConstants.java and the template's config class?

poojabasker20:
Since these properties are optional, they should be present in TemplateConstants.java and the config class. In case the user doesn't set these properties, they should be set to the default values mentioned in the config class.

vanshaj-bhatia:
Got it. LGTM.

vanshaj-bhatia (approved):

PR #535, opened 2023-01-25
feat: Built Python CASSANDRAtoGCS template for Dataproc

saumyasinha-google:


vanshaj-bhatia:
@saumyasinha-google I can see some unwanted files, please remove them from the branch.

saumyasinha-google:
> @saumyasinha-google I can see some unwanted files, please remove them from the branch.

Removed all the unwanted files.

python/dataproc_templates/cassandra/cassandra_to_gcs.py
```diff
@@ -0,0 +1,163 @@
...
+        .config(constants.SQL_EXTENSION, constants.CASSANDRA_EXTENSION)
+        .config(f"spark.sql.catalog.{catalog}", constants.CASSANDRA_CATALOG)
+        .config(f"spark.sql.catalog.{catalog}.spark.cassandra.connection.host",input_host)
+        .getOrCreate())
+ 
+       # Read
+       if(not query):
+           input_data = spark.read.table(f"{catalog}.{input_keyspace}.{input_table}")
+       else:
+           input_data = spark.sql(query)
```
vanshaj-bhatia, line 147: As per the current design, either query or {input_keyspace}.{input_table} would be required for execution. That means, if the user provides query, then {input_keyspace}.{input_table} are not to be provided. This is not handled correctly as {input_keyspace}.{input_table} are set as required properties.

vanshaj-bhatia (changes requested):

python/dataproc_templates/cassandra/cassandra_to_gcs.py
saumyasinha-google: Modified required properties for keyspace and table to be true only when query is None.
Thanks for pointing that out!

vanshaj-bhatia: Please update README as well.

vanshaj-bhatia: @saumyasinha-google What would happen if all three params: table, keyspace and query is provided by the user?

saumyasinha-google: The template will run fine if all three will be provided. However keyspace and table name won't be required properties in such a case but even if they are provided template will run fine.

vanshaj-bhatia: Ideally, it should fail and prompt the user to either provide query or table and keyspace. Please make the necessary changes.

saumyasinha-google: Made the changes.

vanshaj-bhatia (approved):
LGTM

PR #566, opened 2023-02-01
fix: add default values to template.properties, remove duplicate save & unused prop in Wordcount

tanyarw:
1. Cassandra to BQ
2. Cassandra to GCS
3. GCS to JDBC
4. GCS to Spanner
5. JDBC to Spanner


vanshaj-bhatia (changes requested):
I see some other properties like batchInsertSize still have defaults at Config class. Can we also move then to template.properties file?

vanshaj-bhatia (approved):
LGTM!

PR #568, opened 2023-02-02
fix: Issue##485 - Enabled partition level writes in HIVE2GCS template

ritika-neema:
cc: @hhasija - Enabled partition level writes on GCS as a default behaviour, with exception of sql query being provided.

### DIFF
--- a/python/dataproc_templates/hive/hive_to_gcs.py
+++ b/python/dataproc_templates/hive/hive_to_gcs.py
@@ -17,7 +17,9 @@ from logging import Logger
 import argparse
 import pprint
 import sys
-from pyspark.sql import SparkSession, DataFrameWriter
+import re
+from pyspark.sql import SparkSession, DataFrameWriter, DataFrame
+from pyspark.sql.functions import regexp_replace, col, regexp_extract, explode, split, trim
 
 from dataproc_templates import BaseTemplate
 import dataproc_templates.util.template_constants as constants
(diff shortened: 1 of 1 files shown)


python/dataproc_templates/hive/hive_to_gcs.py
```diff
@@ -139,8 +155,17 @@ def run(self, spark: SparkSession, args: Dict[str, Any]) -> None:
...
+        #Get partition columns from Hive
+        partition_cols = self.get_hive_partitions(input_table_name,spark)
+        logger.info(
+                "Method get_hive_partitions: received partition columns as: "
+                f"{partition_cols} for hive table {input_table_name}"
+                )
+
         # Write
         writer: DataFrameWriter = output_data.write.mode(output_mode)
+        if not sql_query.strip() and partition_cols:
```
vanshaj-bhatia, line 167: Why do we need `not sql_query.strip()` in this condition here?

vanshaj-bhatia (changes requested):

python/dataproc_templates/hive/hive_to_gcs.py
ritika-neema: In case a user selects only a few columns when sql_query is used, we might have a situation when partition column is not in selected columns. We could crosscheck columns and then write data onto partitions, but agreed with Hitesh that it overcomplicates the issue. To summarize, we retain partitions by default when moving data from a table directly and not using sql query.

vanshaj-bhatia (approved):
LGTM!

PR #576, opened 2023-02-09
feat: Add DataplexGCStoBQ JUnit test with relevant changes to template 

mugdhapattnaik:


java/bin/start.sh
```diff
@@ -70,6 +70,7 @@ OPT_JARS="--jars=file:///usr/lib/spark/external/spark-avro.jar,${GCS_STAGING_LOC
 OPT_LABELS="--labels=job_type=dataproc_template"
 OPT_DEPS_BUCKET="--deps-bucket=${GCS_STAGING_LOCATION}"
 OPT_CLASS="--class=com.google.cloud.dataproc.templates.main.DataProcTemplate"
+OPT_SPARK_VERSION="--version=1.0"
```
shashank-google, line 73: Why are we doing this ?
Where is git issue for this.
Dataproc Serveless currently supports Spark 2.0, why are we going backwards as default ?
@mugdhapattnaik can you please follow the process:
1. create github issue for each code change explaining details
2. Code "only" against that git issue
3. Raise pull request with the respective git issue linked.

Please remove changes in this file from this pull request and follow the above process.
@hhasija @vanshaj-bhatia FYI

shashank-google (dismissed):

java/bin/start.sh
mugdhapattnaik: Hi Shashank
I had noticed the java integration tests were failing yesterday. The default value of Spark runtime version that was being used was 2.0 and the default Scala version for this Spark version is 2.13. However the templates cannot run with Scala 2.13 because the connectors were meant to run with Scala 2.12. Hitesh mentioned we would be good to handle this change in this PR itself.
Hope that clarifies why this was included here.

I have created new issue now, will add this code change against that issue. [https://github.com/GoogleCloudPlatform/dataproc-templates/issues/577](https://github.com/GoogleCloudPlatform/dataproc-templates/issues/577)

hhasija: sure @shashank-google , we would be going with a separate issue creation onto this.

java/src/test/java/com/google/cloud/dataproc/templates/dataplex/DataplexGCSToBQTest.java
```diff
@@ -0,0 +1,81 @@
...
+    Properties props = PropertyUtil.getProperties();
+    props.setProperty(PROJECT_ID_PROP, "projectID");
+    props.setProperty(DATAPLEX_GCS_BQ_TARGET_DATASET, "dataplex_gcs_to_bq");
+    props.setProperty(
+        DATAPLEX_GCS_BQ_TARGET_ENTITY,
+        "projects/yadavaja-sandbox/locations/us-west1/lakes/dataproc-templates-test-lake/zones/dataplex-gcs-to-bq/entities/dataplex_gcs_to_bq");
+    dataplexGCStoBQTest =
+        new DataplexGCStoBQ(
+            null,
+            "projects/yadavaja-sandbox/locations/us-west1/lakes/dataproc-templates-test-lake/zones/dataplex-gcs-to-bq/entities/dataplex_gcs",
```
vanshaj-bhatia, line 58: Please update the project name to dummy values.

mugdhapattnaik: Have made the requested changes.

hhasija (approved):
LGTM. But, please do update this branch before merging it to main.

PR #580, opened 2023-02-12
feat: Added optional arguments to make DDL compatible with Spark Tables and BQ Translation API

shubhamgoogle:
Changes:
Added Optional Arguments:
 a. HIVE_DDL_CONSIDER_SPARK_TABLES: Will remove "AS SERDE" from the SQL if set to TRUE
 b. HIVE_DDL_TRANSLATION_DISPOSITION: Remove "LOCATION" parameter and "USING FORMAT" if set to TRUE
 
NOTE: using comes when it is a spark table and TRANSLATION API does not recognise it as a HIVE DDL
Example:
>>> spark.sql("show create table employees_partitioned").show(10,False)

|CREATE TABLE `default`.`employees_partitioned` (
  `eid` INT,
  `name` STRING,
  `position` STRING,
  `dept` STRING,
  `part1` INT,
  `part2` STRING)
USING text
PARTITIONED BY (part1, part2)
LOCATION 'hdfs://hive-cluster-m/user/hive/warehouse/partitiondb.db/employee'
TBLPROPERTIES (
  'transient_lastDdlTime' = '1670464514',
  'bucketing_version' = '2')
  
The above SQL gives error if supplied directly to the Translation API.

### DIFF
--- a/python/dataproc_templates/hive/util/README.md
+++ b/python/dataproc_templates/hive/util/README.md
@@ -2,9 +2,7 @@
 
 This is a PySpark utility to extract DDLs of all the tables from a HIVE database using Hive Metastore. Users can use this utility to accelerate their Hive migration journey.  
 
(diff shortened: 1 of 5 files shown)


NiloFreitas (approved):
Hey @shubhamgoogle , I commited some minor changes, LGTM now. Feel free to merge it. Thanks, nice work

PR #581, opened 2023-02-13
fix: add Spark runtime version to fix #577

### LINKED ISSUE #577: [BUG][java][Python] Integration tests failing on main
The default value of Spark runtime version that is being used is 2.0 and the default Scala version for this Spark version is 2.13. However the templates cannot run with Scala 2.13 because the connectors were meant to run with Scala 2.12

mugdhapattnaik:


surjits254 (approved):
LGTM

PR #629, opened 2023-02-21
feat: add azure to bigquery pyspark template 

mokhahmed:


vanshaj-bhatia:
Is there any GitHub issue for this?

shashank-google:
@vanshaj-bhatia request you to please create a new github issue for this.
@hhasija fyi.

Also @mokhahmed will you be willing to create a similar template for Java too ?
So that we are in functional parity on both Java and Python.

README.md
```diff
@@ -62,6 +62,7 @@ Please refer to the [Dataproc Templates (Python - PySpark) README](/python) for
 * [RedshiftToGCS](/python/dataproc_templates/redshift#redshift-to-gcs) (blogpost [link](https://medium.com/google-cloud/exporting-data-from-redshift-to-gcs-using-gcp-dataproc-serverless-and-pyspark-9ab78de11405))
 * [SnowflakeToGCS](/python/dataproc_templates/snowflake#1-snowflake-to-gcs) (blogpost [link](https://medium.com/@varunikagupta96/exporting-data-from-snowflake-to-gcs-using-pyspark-on-dataproc-serverless-363d3bed551b))
 * [TextToBigQuery](/python/dataproc_templates/gcs#text-to-bigquery) (blogpost [link](https://medium.com/google-cloud/dataproc-serverless-pyspark-template-for-ingesting-compressed-text-files-to-bigquery-c6eab8fb6bc9))
+* [AzureToBigQuery](/python/dataproc_templates/azure#azure-to-bigquery)
```
vanshaj-bhatia, line 65: Please move it to the lexicographical order relative to other templates.

vanshaj-bhatia, line 65: Add the same to Python README [Dataproc Templates (Python - PySpark) README](https://github.com/GoogleCloudPlatform/dataproc-templates/blob/26f22d0b3de819d384c7f89dc903f3955e8b20f2/python).

python/dataproc_templates/azure/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+* `azure.storage.account`: Azure Storage Account  
+* `azure.container.name`:  Azure Container Name
+* `azure.sas`: Azure SAS Token
+* `azure.bigquery.input.location`: Azure blob of the input files (format: `"wasbs://{azure.container.name}@{azure.storage.account}.blob.core.windows.net/{input_location}"`)
+* `azure.bigquery.output.dataset`: BigQuery dataset for the output table
+* `azure.bigquery.output.table`: BigQuery output table name
+* `azure.bigquery.input.format`: Input file format (one of: avro,parquet,csv,json)
+* `azure.bigquery.temp.bucket.name`: Temporary bucket for the Spark BigQuery connector
+* `azure.bigquery.output.mode`: Output write mode (one of: append,overwrite,ignore,errorifexists)(Defaults to append)
```
vanshaj-bhatia, line 17: Bit unconvinced from the nomenclature we are following here. Using **Azure** alone in the template name and the template properties doesn't make sense as we are exporting data specifically from **Azure Blob Storage**.

python/dataproc_templates/azure/README.md
```diff
@@ -0,0 +1,80 @@
+# Azure To BigQuery
```
vanshaj-bhatia, line 1: Please update to Azure Blob

vanshaj-bhatia (changes requested):

python/dataproc_templates/azure/README.md
mokhahmed: i believe azure convention is Azure Storage Account not Azure Blob Storage Account and the same for Azure Container Name. should we rename them?

vanshaj-bhatia: So, Using Azure Storage Account also would not be correct here. As a Storage Account can support services for Blob, Tables, Queues, ADLS etc. 
Since, we are dealing with only blob storage here, we should be good with using _Azure Blob Storage_ in the template name , and _azure.blob.bigquery*_ in the templateProperty.

vanshaj-bhatia:
Please create unit test cases for this template.

vanshaj-bhatia (approved):
LGTM!

PR #634, opened 2023-02-23
fix: Added ValidateInput method to CassandraToBQ

saumyasinha-google:


java/src/main/java/com/google/cloud/dataproc/templates/databases/CassandraToBQ.java
```diff
@@ -80,5 +82,24 @@ public void runTemplate() throws StreamingQueryException, TimeoutException {
         .save();
   }
 
-  public void validateInput() {}
+  public void validateInput() {
+    if (StringUtils.isAllBlank(config.getHost())
+        || StringUtils.isAllBlank(config.getBqLocation())
+        || StringUtils.isAllBlank(config.getTemplocation())) {
```
vanshaj-bhatia, line 88: Cassandra to BQ already has Jakarta validations in place using the Config class. We need not have to perform validations again.

vanshaj-bhatia (changes requested):

java/src/main/java/com/google/cloud/dataproc/templates/databases/CassandraToBQ.java
saumyasinha-google: I thought the same, but the issue was created to add validation code in the validateInput method.
Should I close all the validateInput issues on templates which already have Jakarta validations?
(There are a few more open issues for Spanner to GCS, Redshift to GCS etc)

hhasija: Actually @saumyasinha-google , the idea was to keep the Jakarta validation framework and the calling of this framework should be done via validateInput method. So that ways, both the agenda would be fulfilled

1. We would be having Jakarta validation framework in our templates.
2. We would be having validateInput method, to check for validations.

saumyasinha-google: Added Jakarta validations in validateInput method.

java/src/main/java/com/google/cloud/dataproc/templates/databases/CassandraToBQ.java
```diff
@@ -80,5 +81,7 @@ public void runTemplate() throws StreamingQueryException, TimeoutException {
         .save();
   }
 
-  public void validateInput() {}
+  public void validateInput() {
+    ValidationUtil.validateOrThrow(config);
```
poojabasker20, line 85: Calling ValidationUtil.validateOrThrow(config) from validateInput() causes the unit test to fail since the logic is such that it expects the validation exception to be thrown in of().

Calling validateInput() from of() instead of runTemplate() would resolve this issue.

poojabasker20 (changes requested):

vanshaj-bhatia:
@saumyasinha-google please take latest changes (pertaining to calling validateInput() from DataprocTemplates.java) from main branch

tanyarw (approved):
LGTM!

vanshaj-bhatia (approved):

PR #643, opened 2023-02-27
fix: 635 java write mode should be configurable via templateproperties

tanyarw:
1. BQ to GCS
2. GCS to BQ
3. S3 to BQ


vanshaj-bhatia (approved):
LGTM!

PR #646, opened 2023-02-28
fix: 600 java template gcstospanner empty validateinput method

saumyasinha-google:


java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCSToSpanner.java
```diff
@@ -82,5 +81,7 @@ public void write(Dataset<Row> dataset) {
         .save();
   }
 
-  public void validateInput() {}
+  public void validateInput() {
+    ValidationUtil.validateOrThrow(config);
```
poojabasker20, line 85: Calling ValidationUtil.validateOrThrow(config) from validateInput() causes the unit test to fail since the logic is such that it expects the validation exception to be thrown in of().

Calling validateInput() from of() instead of runTemplate() would resolve this issue.

poojabasker20 (changes requested):

vanshaj-bhatia:
@saumyasinha-google please take latest changes (pertaining to calling validateInput() from DataprocTemplates.java) from main branch

hhasija (approved):
LGTM

poojabasker20 (approved):

PR #649, opened 2023-03-01
fix: 598 java template spannertogcs empty validateinput method

tanyarw:


tanyarw:
Will update after #747 is merged

PR #659, opened 2023-03-06
fix: partition write fix in HIVE2GCS template

ritika-neema:
Fix for Issue#485 partition writes in HIVE2GCS template


vanshaj-bhatia (approved):
LGTM!

PR #666, opened 2023-03-07
fix: add runtime_config_version to notebooks

### LINKED ISSUE #665: [Notebook] Fix errors due to new runtime version with Scala 2.13 instead of 2.12
The new default runtime of Dataproc Serverless (version 2) uses Scala 2.13, which makes some dependencies, compiled using Scala 2.12 to throw errors, like Caused by: java.lang.NoClassDefFoundError: scala/Serializable.

For now, we need to set the specific runtime version when submitting jobs, to use runtime version 1.

We need to update all notebooks to pass the **runtime_config_version** when submitting the job. Ref: https://google-cloud-pipeline-components.readthedocs.io/en/google-cloud-pipeline-components-1.0.6/google_cloud_pipeline_components.experimental.dataproc.html

Runtime versions: https://cloud.google.com/dataproc-serverless/docs/concepts/versions/spark-runtime-versions

NiloFreitas:
This PR adds the runtime_config_version property = "1.1" to the DataprocPySparkBatchOp, when submitting dataproc serverless jobs from the notebooks, to fix the error caused by the new default runtime, incompatible with the dependencies that we are using, which were compiled against Scala 2.12.

fix #665

### DIFF
--- a/notebooks/generic_notebook/vertex_pipeline_pyspark.ipynb
+++ b/notebooks/generic_notebook/vertex_pipeline_pyspark.ipynb
@@ -4,7 +4,11 @@
    "cell_type": "code",
    "execution_count": null,
    "id": "d2c62b00-03d1-4d9e-beb2-f9961452338a",
-   "metadata": {},
+   "metadata": {
+    "pycharm": {
+     "name": "#%%\n"
+    }
+   },
    "outputs": [],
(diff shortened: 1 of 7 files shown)


surjits254 (approved):
LGTM

Issue #667, opened 2023-03-07
[Enhancement] Upgrade HBASE/BIGTABLE templates to using dataproc version 1.1


surjits254:
currently, all python templates are using version 1.1 except for HBASETOGCS AND GCSTOBIGTABLE template which is still using 1.0.29. Logic is defined in start.sh

Test and fix remaining 2 templates above and remove the interim logic in start.sh to make all templates use version 1.1

surjits254:
This is not required. Newer version will run on 1.2

PR #668, opened 2023-03-07
fix: Fix 591 enhancement ugrade dataproc version to latest for python

surjits254:
this upgrades all python templates to use version 1.1 except HBASETOGCS and GCSTOBIGTABLE. Issue #667 is created to upgrade remaining 2 templates to 1.1 version


NiloFreitas (approved):
LGTM but need to be addressed later why 1.1 does not work with hbase and bigtable. Probably we need to wait the dependencies related to them and also others like the spark bigquery connector to be updated so we can use version 2 actually

PR #673, opened 2023-03-10
fix: 601 java template generaltemplate empty validateinput method

tanyarw:


java/src/main/java/com/google/cloud/dataproc/templates/general/GeneralTemplate.java
```diff
@@ -84,9 +85,10 @@ public static GeneralTemplate of(String... args) {
     } catch (IOException e) {
       throw new IllegalArgumentException("Could not load config yaml", e);
     }
+    GeneralTemplate generalTemplate = new GeneralTemplate(config);
```
shashank-google, line 88: @hhasija 
Looks like we are doing validation twice here.
I think code needs to be refactored so that DataProcTemplate.main will call BaseTemplate.validate().

Also from all the individual templates like this one, we got to remove call to validate() from of method or runTemplate() method. The junit tests are required to be updated accordingly as well.

shashank-google (changes requested):

java/src/main/java/com/google/cloud/dataproc/templates/general/GeneralTemplate.java
tanyarw: Made changes

tanyarw:
Will update after #747 is merged

vanshaj-bhatia (approved):
LGTM!

shashank-google (approved):

PR #693, opened 2023-03-16
fix: 661 assess and fix incorrect handling of try catch for all java and python templates

tanyarw:
Java templates:
1. BQ To GCS
2. GCS To BQ
3. PubSub to BT
4. PubSub to BQ
5. PubSub to GCS
6. S3 to BQ


vanshaj-bhatia (approved):
LGTM!

PR #694, opened 2023-03-16
Hive ddl extractor enhancement

shubhamgoogle:
feat: Added optional arguments to make DDL compatible with Spark Tables and BQ Translation API


NiloFreitas (approved):
LGTM

PR #747, opened 2023-04-18
fix: 739 java refactor code to avoid duplicate validation

tanyarw:
1. Called `validateInput()` from main
2. Removed validateInput call from all templates `runTemplate()` or `of()`
3. Updated unit tests to assert with `template.validateInput()`


vanshaj-bhatia (commented):
LGTM! Great work @tanyarw...

@hhasija could you please have a quick look as well?

PR #778, opened 2023-05-11
fix: updating read me files with cassandra to gcs blogpost link

hhasija:


hhasija:
merged as a part of  https://github.com/GoogleCloudPlatform/dataproc-templates/pull/761

PR #782, opened 2023-05-12
feat: 774 create a parameterize script for oracletobigquery notebook

tanyarw:


nj1973 (commented):
The code looks good to me. I did have a couple of issues testing though.

```
ModuleNotFoundError: No module named 'google'
```

I think I ran into this because `pip3 install --upgrade google-cloud-pipeline-components kfp --user -q` fails with:
```
ERROR: Can not perform a '--user' install. User site-packages are not visible in the virtualenv.
```


I ran these manually:
```
python3 -m pip install google.cloud.aiplatform
python3 -m pip install --upgrade google-cloud-pipeline-components kfp
```

The next problem was:
```
ModuleNotFoundError: No module named 'pandas'
```

I ran this manually, it's not clear why I had to do this:
```
python3 -m pip install pandas
```

After this the code ran through successfully.

I wonder if we need to add any of the items above to the requirements file?

tanyarw:
> 

That's a good catch! Some libraries aren't being installed through the notebook. I'll update requirements file.

What about the `--user` install? Should this be fixed in the notebook or added as a requirement?

nj1973:
> > 
> 
> @nj1973 What about the `--user` install? Should this be fixed in the notebook or added as a requirement?

I don't actually know which Python the notebooks use when used interactively. I presume it is not the system Python.
My feeling is that we should not include `--user` but I wonder if @NiloFreitas has better knowledge than me?

tanyarw:
> > > 
> > 
> > 
> > @nj1973 What about the `--user` install? Should this be fixed in the notebook or added as a requirement?
> 
> I don't actually know which Python the notebooks use when used interactively. I presume it is not the system Python. My feeling is that we should not include `--user` but I wonder if @NiloFreitas has better knowledge than me?

The kernel to be used can be mentioned in this [function](https://github.com/GoogleCloudPlatform/dataproc-templates/pull/782/files#diff-d9fb22c5cdbae429a40c3501a1c53d30f242f0e4b84a4f45a58808bcf6096ff0:~:text=pm.execute_notebook,) by having a `kernel_name` parameter.
[Doc link](https://papermill.readthedocs.io/en/latest/reference/papermill-workflow.html#:~:text=kernel_name%20(str)%20%E2%80%93%20Name%20of%20kernel%20to%20execute%20the%20notebook%20against.) for the same

nj1973 (approved):
Looks good to me.

PR #798, opened 2023-05-26
Fix: Prevents integration tests from overwriting jar file when multiple PRs are raised

poojabasker20:


java/.ci/Jenkinsfile
```diff
@@ -803,6 +814,13 @@ pipeline {
     				notifyBackToNormal: 'true'
                 }
             }
+            catchError {
+                sh '''
+                if [ $GIT_BRANCH_LOCAL != "main" ];then
+                    gsutil rm -r $GCS_STAGING_LOCATION 2> /dev/null || true
+                fi
```
vanshaj-bhatia, line 821: Would the catchError block only run when the pipeline fails? Ideally we want to remove the GCS location in successful runs as well to avoid cluterring the GCS bucket.

poojabasker20: No, this block will run every time. Please refer - https://dataproctemplatesci.com/job/dataproc-templates-build/job/ad-hoc/job/integration-tests-java/313/console

vanshaj-bhatia (approved):
LGTM

PR #807, opened 2023-06-05
fix: 806 code refactoring to remove code duplication

tanyarw:
- De-duplicate`get_env_var()` function that is being used repeatedly in every parameterize file. [Refer here](https://github.com/GoogleCloudPlatform/dataproc-templates/blob/main/notebooks/mysql2spanner/MySqlToSpanner_parameterize_script.py#L140)
- Implement papermill logging along with `get_common_args()` function

### DIFF
--- a/notebooks/README.md
+++ b/notebooks/README.md
@@ -55,7 +55,6 @@ export GCS_STAGING_LOCATION=<gs://bucket-name>
 export SUBNET=<subnet>
 
 python run_notebook.py --script=<NOTEBOOK_NAME> \
-   --log_level=<LOG_LEVEL> \
    --notebook.paramter1="<>" \
    --notebook.paramter2="<>"
 ```
--- a/notebooks/mysql2spanner/README.md
+++ b/notebooks/mysql2spanner/README.md
@@ -81,7 +81,6 @@ python run_notebook.py --script=MYSQLTOSPANNER \
 python run_notebook.py --script=MYSQLTOSPANNER --help
 
 usage: run_notebook.py [-h]
-        [--output.notebook OUTPUT.NOTEBOOK]
         --mysql.host MYSQL_HOST
         [--mysql.port MYSQL_PORT]
         --mysql.username MYSQL_USERNAME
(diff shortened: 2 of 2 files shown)


notebooks/parameterize_script/util/argument_parsing.py
```diff
@@ -55,3 +55,34 @@ def get_script_name(args: Optional[Sequence[str]] = None) -> ScriptName:
...
+            By default, command line arguments will be used.
+    Returns:
+        str: The value of the --log_level argument
+    """
+
+    parser: argparse.ArgumentParser = argparse.ArgumentParser(
+        add_help=False
+    )
+
+    parser.add_argument(
```
nj1973, line 75: I find this a bit confusing. We've already added our arguments and parsed them in `script_instance.parse_args()`, so it doesn't feel right to be adding more arguments and re-parsing here.
Should `MySqlToSpannerScript.parse_args` add global/common arguments instead?
Something like:
```
def parse_args(args: Optional[Sequence[str]] = None) -> Dict[str, Any]:
    parser = argparse.ArgumentParser()
    add_common_args(parser)
    parser.add_argument(
            f'--{constants.OUTPUT_NOTEBOOK_ARG}',
...
```

The the line in `run_script()` would be:
```
logging.basicConfig(level=args.log_level, format="%(message)s")
```

This is just an idea, feel free to reject it. It just seems odd to keep re-parsing the arguments. Same goes for `get_script_name()`, that would be added and validated parse_args too.

tanyarw: common_args() is a nice idea, will implement that

notebooks/parameterize_script/util/argument_parsing.py
```diff
@@ -55,3 +55,34 @@ def get_script_name(args: Optional[Sequence[str]] = None) -> ScriptName:
...
+        '--log_level',
+        dest='log_level',
+        type=str,
+        required=False,
+        default="INFO",
+        choices=["NOTSET","DEBUG","INFO","WARNING","ERROR","CRITICAL"],
```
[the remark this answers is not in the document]
tanyarw: Yep, alright

nj1973 (approved):
LGTM

PR #808, opened 2023-06-08
docs: removed PubSubLiteToGCS from README files

vanshaj-bhatia:
Removing PubSubLiteToGCS python template from READMEs as the template is not working correctly.


hhasija (approved):
LGTM

PR #862, opened 2023-10-17
feat: added support for running python templates in CLUSTER mode

### LINKED ISSUE #849: [Python] Add support to run Python templates on an existing cluster
Similar to Java's bin/start.sh. Python templates should also expect two environment variables:

- JOB_TYPE
- CLUSTER

Run the templates on a existing cluster, is the JOB_TYPE is set to "CLUSTER".

Make the respective changes in the Python's bin/start.sh

vanshaj-bhatia:
closes #849


surjits254 (approved):
LGTM

PR #869, opened 2023-10-30
docs: Add recommended jar versions to run MongoToGCS template

SatpreetMakhija:
MongoToGCS template does not work for some versions of required jars. Therefore, adding recommended jar versions for the template. This has been discussed with  @surjits254.


ankuljain09:
@SatpreetMakhija We can close this PR for now as we have another PR created from a feature branch

PR #936, opened 2024-08-25
feat: Provide optional parameter service account name for serverless jobs

rajc242:
feat: Provide optional parameter service account name while submitting serverless jobs


surjits254 (approved):
LGTM

PR #957, opened 2024-09-18
feat: Spanner JDBC driver can use postgresql dialect

rajc242:
1. Upgraded spanner jdbc driver from `2.4.2` to `2.6.0` to support single jdbc driver with googlesql and postgresql dialect. Ref: [Documentation](https://github.com/googleapis/java-spanner-jdbc/blob/main/CHANGELOG.md#260-2022-02-24)
2. Added PostgresJDBCDialect to map spark datatypes with postgres datatypes
3. Fix jenkins test case
4. Update README.md files for all affected templates
5. Backword compatibility remains as it is as all affected templates are by default using GoogleSQL dialect


java/src/main/java/com/google/cloud/dataproc/dialects/PostgresJDBCDialect.java
```diff
@@ -0,0 +1,90 @@
...
+    } else {
+      return Option.empty();
+    }
+  }
+
+  @Override
+  public Option<DataType> getCatalystType(
+      int sqlType, String typeName, int size, MetadataBuilder md) {
+    if (sqlType == Types.NUMERIC || sqlType == Types.DECIMAL)
+      return Option.apply(DecimalType.apply(38, 9));
```
shashank-google, line 87: Looking at the Spanner [documentation](https://cloud.google.com/spanner/docs/working-with-numerics#postgresql-numeric) it does not support precision and scale.
Are we sure we need `DecimalType.apply(38, 9)` ??
Can we please test this work by using a source data which has decimal type ?

shashank-google (changes requested):

java/src/main/java/com/google/cloud/dataproc/dialects/PostgresJDBCDialect.java
rajc242: Validated and added test case with decimal type to avoid any future break due to unintentional code changes

shashank-google (approved):

PR #963, opened 2024-09-30
feat: Initial commit for Spark runtime update to 1.2

rajc242:
1. Serverless dataproc version migration from `1.1` to `1.2`
2. Maintained maven compile version to JDK11
3. Change bigquery format to more enterprise standard
4. Upgrade existing libraries to match with supported version of `1.2`
5. Added more relocations in pom.xml to support spanner jdbc driver


java/pom.xml
```diff
@@ -35,83 +35,313 @@
     </license>
   </licenses>
   <properties>
-    <java.version>1.8</java.version>
+    <!-- https://cloud.google.com/dataproc-serverless/docs/concepts/versions/spark-runtime-versions#spark_runtime_version_12 -->
+    <!-- Java & Scala Version Aligned with Dataproc Serverless Spark Runtime 1.2 -->
+    <java.version>17</java.version>
```
shashank-google, line 40: First we need to submit the upgrade to version java 11 + dataproc serverless 1.2

java/bin/start.sh
```diff
@@ -64,10 +64,10 @@ if [ -z "$SKIP_BUILD" ]; then
 
 fi
 
-OPT_SPARK_VERSION="--version=1.1"
+OPT_SPARK_VERSION="--version=1.2"
 OPT_PROJECT="--project=${GCP_PROJECT}"
 OPT_REGION="--region=${REGION}"
-OPT_JARS="--jars=file:///usr/lib/spark/external/spark-avro.jar,${GCS_STAGING_LOCATION}/${JAR_FILE}"
```
shashank-google, line 70: This will be a breaking change for UI, @surjits254 can you please verify ?
@rajc242  Is this really needed or it is more like a clean up effort ?

java/src/main/java/com/google/cloud/dataproc/templates/config/GeneralTemplateConfig.java
```diff
@@ -40,7 +40,8 @@
 public class GeneralTemplateConfig {
 
   @NotEmpty() private Map<String, @Valid InputConfig> input = ImmutableMap.of();
-  @NotNull private Map<String, @Valid QueryConfig> query = ImmutableMap.of();;
+  @NotNull private Map<String, @Valid QueryConfig> query = ImmutableMap.of();
+  ;
```
shashank-google, line 44: extra semicolon

java/src/main/java/com/google/cloud/dataproc/templates/config/OutputConfig.java
```diff
@@ -26,7 +26,8 @@ public class OutputConfig {
 
   @NotBlank private String format;
   private String path;
-  @NotNull private Map<String, String> options = ImmutableMap.of();;
+  @NotNull private Map<String, String> options = ImmutableMap.of();
+  ;
```
shashank-google, line 30: extra semicolon

java/bin/start.sh
rajc242: Spark runtime 1.2 is failing with message that file not found exception.

shashank-google: Oh ok. What happens if we remove reference of this jar in both places i.e. start.sh and jenkins file ?
Does it run with both v1.1 and v1.2 ?

rajc242: We didn't change anything for `1.1` so it is working as expected but `1.2` is not working. I explicitly provided jar path through `JARS` option and it is working fine.

java/pom.xml
rajc242: I am in support of JDK 17 instead of JDK 11. I have seen lots of people have been already migrated to JDK 17. I noticed lots of questions on stackoverflow for Dataproc 2.x

Even I am in support of starting to think about dataproc 2.x migration

shashank-google: We can't go to java 17 directly, since we got to support both S8S v1.1 and S8S v1.2 for some time (until the UI and downstream is moved to 1.2).
Therefore we need to change to Java 11

rajc242: I am trying to downgrade it with Java 11.

shashank-google (approved):

PR #964, opened 2024-10-06
ci: 906 UI templates integration tests and cicd

surjits254:
integration tests for UI.
ToDo after merge:
1. add to each PR and release
2. schedule to run daily


java/.ci/UiJenkinsfile
```diff
@@ -0,0 +1,468 @@
...
+                                    --templateProperty jdbctogcs.sql="select * from test.employee" \
+                                    --templateProperty jdbctogcs.write.mode="overwrite"
+                            '''
+                        }
+                    }
+                }
+            }
+        }
+    }
+    // post {
```
vanshaj-bhatia, line 454: We can uncomment it and add a conditinal statement similar to https://github.com/GoogleCloudPlatform/dataproc-templates/blob/28c155fbf1e3322de9e9fbe376b1b5078c565042/java/.ci/Jenkinsfile#L1099

Which would trigger notifications only when it runs for main branch.

vanshaj-bhatia:
Can we add a step to build jar from the current branch code and use that for running gcloud commands.

vanshaj-bhatia (changes requested):
GCS_DESTINATION is current date based. My suggestion is to use uuidgen, to avoid overriding jar files if multiple PRs are raised at same time.

vanshaj-bhatia (approved):
LGTM!

PR #1000, opened 2024-10-21
feat: BigQuery_To_JDBC

hhasija:


rajc242 (changes requested):
Could you please provide validation through validation class as [JDBCToGCSConfig](https://github.com/GoogleCloudPlatform/dataproc-templates/blob/main/java/src/main/java/com/google/cloud/dataproc/templates/jdbc/JDBCToGCSConfig.java)

We are trying to standardize validation through config class for new templates.

java/src/main/java/com/google/cloud/dataproc/templates/bigquery/BigQueryToJDBC.java
```diff
@@ -0,0 +1,92 @@
...
+
+  @Override
+  public void runTemplate() {
+
+    SparkSession spark = SparkSession.builder().appName("BigQuery to JDBC").getOrCreate();
+
+    // Set log level
+    spark.sparkContext().setLogLevel(sparkLogLevel);
+
+    Dataset<Row> inputData = spark.read().format(SPARK_READ_FORMAT_BIGQUERY).load(inputTableName);
```
rajc242, line 59: Can we please enable below two properties which will use storage read api with faster performance. Reference [documentation](https://github.com/GoogleCloudDataproc/spark-bigquery-connector?tab=readme-ov-file#reading-data-from-a-bigquery-query)

```java
spark.conf.set("viewsEnabled","true")
spark.conf.set("materializationDataset","<dataset>")
```

java/src/main/java/com/google/cloud/dataproc/templates/bigquery/README.md
```diff
@@ -34,4 +34,46 @@ bin/start.sh \
...
+export GCP_PROJECT=<gcp-project-id>
+export REGION=<region>
+export SUBNET=<subnet> 
+export GCS_STAGING_LOCATION=<gcs-staging-bucket-folder>
+export HISTORY_SERVER_CLUSTER=<history-server>
+
+bin/start.sh \
+-- --template BIGQUERYTOJDBC \
+--templateProperty project.id=<gcp-project-id> \
+--templateProperty bigquery.jdbc.input.table=<projectId:datasetId.tableName> \
```
rajc242, line 54: Ideally connector is accepting query as well table name. If we can provide one example with query then it would be great. Example if table has millions of row and anyone wants to query and load only subset then user might pass query instead of table name.

hhasija: Changes done. Please review it once.

java/src/main/java/com/google/cloud/dataproc/templates/bigquery/BigQueryToJDBC.java
```diff
@@ -0,0 +1,81 @@
...
+
+  @Override
+  public void runTemplate() {
+
+    // Create a SparkConf object and set configurations
+    SparkConf conf =
+        new SparkConf()
+            .setAppName("BigQuery to JDBC")
+            .set("spark.sql.viewsEnabled", "true")
+            .set("spark.sql.materializationDataset", "<dataset>");
```
rajc242, line 55: Actually we required dataset from users project. Ideally it should be configuration which will be passed from outside as template property.

java/README.md
```diff
@@ -6,6 +6,7 @@
 Please refer to the [Dataproc Templates (Java - Spark) README](java/README.md) for more information.
 
 * [BigQueryToGCS](src/main/java/com/google/cloud/dataproc/templates/bigquery#BigQuery-To-GCS)
```
rajc242, line 8: We also need to add this at main home page [READEME.md](https://github.com/GoogleCloudPlatform/dataproc-templates?tab=readme-ov-file)

rajc242 (changes requested):

java/src/main/java/com/google/cloud/dataproc/templates/bigquery/BigQueryToJDBC.java
hhasija: Made changes. Please review once.

rajc242 (changes requested):
Please could you add one execution in [Jenkins](https://github.com/GoogleCloudPlatform/dataproc-templates/blob/main/java/.ci/Jenkinsfile) file for validation.

hhasija:
> Could you please provide validation through validation class as [JDBCToGCSConfig](https://github.com/GoogleCloudPlatform/dataproc-templates/blob/main/java/src/main/java/com/google/cloud/dataproc/templates/jdbc/JDBCToGCSConfig.java)
> 
> We are trying to standardize validation through config class for new templates.

Changes Done. Please review it once.

hhasija:
> Please could you add one execution in [Jenkins](https://github.com/GoogleCloudPlatform/dataproc-templates/blob/main/java/.ci/Jenkinsfile) file for validation.

Changes done. Please do review it once.

rajc242 (approved):
LGTM

PR #1014, opened 2024-11-02
fix: adding csv support for gcs to spanner template

hhasija:


java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCSToSpanner.java
```diff
@@ -54,10 +54,18 @@ public void runTemplate() {
...
-          spark.read().format(config.getInputFormat()).load(config.getInputLocation());
-
+      Dataset<Row> dataset;
+      if ("csv".equalsIgnoreCase(config.getInputFormat())) {
+        dataset =
+            spark
+                .read()
+                .format(config.getInputFormat())
+                .option("header", "true")
+                .option("inferSchema", "true")
```
rajc242, line 64: Here we are infering schema. I think I am okay but we should put in note that spark selected datatypes to be matched with Spanner table data types otherwise there can be error. I mean we are relying on spark that spark will generate perfect schema. Otherwise all good. LGTM!

rajc242 (approved):

hhasija:
> @hhasija can you also update the readme?

Done

java/src/main/java/com/google/cloud/dataproc/templates/gcs/GCSToSpanner.java
hhasija: Done

PR #1016, opened 2024-11-03
fix: jenkins integration tests optimisation

hhasija:


vanshaj-bhatia:
Can it be done at GitHub Actions level? Something like: If only readme related changes are done, do not run the int tests at all.

hhasija:
> Can it be done at GitHub Actions level? Something like: If only readme related changes are done, do not run the int tests at all.

Not sure @vanshaj-bhatia . As I tried finding that out but did not got anything on that level to handle.

rajc242 (approved):
LGTM

PR #1024, opened 2024-11-13
fix: Jenkins optimisation update

hhasija:


rajc242 (changes requested):
Can we cover up requirements.txt file for python which is like pom.xml 

If we make changes there then we would like to test all python test cases.

hhasija:
> Can we cover up requirements.txt file for python which is like pom.xml
> 
> If we make changes there then we would like to test all python test cases.

Done

rajc242 (approved):
LGTM

PR #1037, opened 2025-03-31
fix : Support multiple tables

hhasija:


vanshaj-bhatia (approved):
LGTM!

python/dataproc_templates/jdbc/jdbc_to_bigquery.py
```diff
@@ -180,35 +180,50 @@ def run(self, spark: SparkSession, args: Dict[str, Any]) -> None:
...
-
-        # Write
-        input_data.write \
-                .format(constants.FORMAT_BIGQUERY) \
-                .option(constants.TABLE, big_query_dataset + "." + big_query_table) \
-                .option(constants.GCS_BQ_TEMP_BUCKET, bq_temp_bucket) \
-                .mode(output_mode) \
-                .save()
+        source_tables_list = input_jdbc_table.split(";")
+        target_tables_list = big_query_table.split(";")
```
surjits254, line 191: instead of list can we accept dictionary for source:target.
Keep these 2 variables same to avoid breaking change but add another variable which accepts a key:value of source:target mapping so users do not have to worry about order

surjits254 (changes requested):

python/dataproc_templates/jdbc/jdbc_to_bigquery.py
hhasija: Actually @surjits254 if we would do that, then we need to change our older pattern of accepting the input parameters. Below mentioned are the two older parameters :-

--jdbc.bigquery.input.table='test.demo' 
--jdbc.bigquery.output.table='jdbctobq'

Thus, without changing any of the older support we need to come up with some solution. Hence, figured out this way. Do you think so we should change the older format to dictionary ?

PR #1072, opened 2025-08-12
Feat: Added secret manager for JDBC url

uriarriaga:


surjits254:
Discussed with @uriarriaga to create PR from dataproc-templates branch instead of forked branch. As integration test runs are not supported from forked branch.

PR #1112, opened 2026-04-28
docs: Update README with archiving notice and template info

shashank-google:
Added note about repository archiving and template usage.


sundar-mudupalli-work (approved):
Thank you.

